PR #6, opened 2023-09-06
feat/conversation app api

Dovchik:
- Covers [App](https://developers.sinch.com/docs/conversation/api-reference/conversation/tag/App/) endpoints.
- Http refactor.
- Typos

### DIFF
--- a/src/Sinch/Conversation/Apps/Credentials/BasicAuthCredentials.cs
+++ b/src/Sinch/Conversation/Apps/Credentials/BasicAuthCredentials.cs
@@ -1,0 +1,44 @@
+﻿using System.Text;
+
+namespace Sinch.Conversation.Apps.Credentials
+{   /// <summary>
+    ///     It consists of a username and a password.
+    /// </summary>
+    public sealed class BasicAuthCredential
+    {
+        /// <summary>
+        ///     Basic auth password.
+        /// </summary>
+#if NET7_0_OR_GREATER
+        public required string Password { get; set; }
+#else
+        public string Password { get; set; }
+#endif
+        
+
+        /// <summary>
+        ///     Basic auth username.
+        /// </summary>
+#if NET7_0_OR_GREATER
+        public required string Username { get; set; }
+#else
+        public string Username { get; set; }
+#endif
+        
+
+        /// <summary>
+        ///     Returns the string presentation of the object
+        /// </summary>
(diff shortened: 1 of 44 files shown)


src/Sinch/Conversation/Apps/Credentials/BasicAuthCredentials.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        ///     Returns the string presentation of the object
+        /// </summary>
+        /// <returns>String presentation of the object</returns>
+        public override string ToString()
+        {
+            var sb = new StringBuilder();
+            sb.Append("class BasicAuthCredential {\n");
+            sb.Append("  Password: ").Append(Password).Append("\n");
```
JPPortier, line 37: Should we offer this critical information ?
This mean we could found clear password in any layer using a print/tostring onto such instance (log files, ...)

src/Sinch/Core/Utils.cs
```diff
@@ -72,7 +72,7 @@ public static T ParseEnum<T>(string value)
                 }
             }
 
-            throw new InvalidOperationException($"Failed to parse {nameof(T)} enum");
+            throw new InvalidOperationException($"Failed to parse {enumType.Name} enum");
```
JPPortier, line 75: Does this mean any unknown value will trigger an exception ?
My thoughts from https://github.com/sinch/sinch-sdk-internal-docs/blob/main/ADR/0005-enums.md was we should be able to handle new values *dynamically*: 
> User is able to create new values if needed.
Deserialization is able to create new instances if needed.

Currently, we will have different behavior between `.net` and `java` SDKs (https://github.com/sinch/sinch-sdk-java-proto/pull/2)

src/Sinch/Conversation/Apps/Credentials/BasicAuthCredentials.cs
Dovchik: Good point. Will be removed.

src/Sinch/Core/Utils.cs
Dovchik: Yeah, this is intended. Currently I want it to be left as it is. We can revisit it latter nonetheless.

JPPortier (approved):

650elx (approved):

PR #8, opened 2023-09-20
Feat/loose enum handling

Dovchik:
Enums are too strict and can throw runtime error if API enum values where extended, but SDK is behind the update. 
Record provide the ability to use any enum string and provide access to predefined values as `static readonly` fields.

### DIFF
--- a/src/Sinch/Numbers/NumberPattern.cs
+++ b/src/Sinch/Numbers/NumberPattern.cs
@@ -26,7 +26,7 @@ namespace Sinch.Numbers
             list.Add(new KeyValuePair<string, string>("numberPattern.pattern", Pattern));
             if (SearchPattern is not null)
                 list.Add(new KeyValuePair<string, string>("numberPattern.searchPattern",
-                    SearchPattern.Value.ToUpperInvariant()));
+                    SearchPattern.Value));
 
             return list;
         }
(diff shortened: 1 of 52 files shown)


src/Sinch/Numbers/NumberPattern.cs
```diff
@@ -18,37 +18,40 @@ public class NumberPattern
...
         /// </summary>
-        public SearchPattern? SearchPattern { get; set; }
+        public SearchPattern SearchPattern { get; set; }
 
         internal IEnumerable<KeyValuePair<string, string>> GetQueryParamPairs()
         {
             var list = new List<KeyValuePair<string, string>>();
             list.Add(new KeyValuePair<string, string>("numberPattern.pattern", Pattern));
-            if (SearchPattern.HasValue)
+            if (SearchPattern is not null)
```
JPPortier, line 27: Not seeing from OAS file this field is optional if `numberPattern.pattern` is set.
Side note, not related to this PR content but for this file:
https://github.com/sinch/sinch-sdk-dotnet/blob/b1cdc555e361b34a8f71639969302fff4fb49819/src/Sinch/Numbers/NumberPattern.cs#L50
There is an inconsistency with doc and OAS vs API implementation.
Current API implementation is using the `CONTAINS` keyword (not `CONTAIN`
When using `CONTAIN`, server will return:
```javascript
"error": {
  "code": 400,
  "message": "parsing field \"search_pattern\": \"CONTAIN\" is not a valid value", 
  "status": "INVALID_ARGUMENT",
  "details": [] 
 }
}
``` 
Ticket opened for API: https://tickets.sinch.com/browse/NUM-5022

Dovchik: They are both optional and independent

JPPortier (approved):

PR #10, opened 2023-10-17
Feat/verification api verification

Dovchik:
https://developers.sinch.com/docs/verification/api-reference/verification/tag/Verifications/ part of api. For additional tests I want to utilize doppleganger latter


src/Sinch/SinchClient.cs
```diff
@@ -100,7 +132,7 @@ public class SinchClient : ISinch
             Numbers = new Numbers.Numbers(projectId, new Uri("https://numbers.api.sinch.com/"),
                 _loggerFactory, httpCamelCase);
 
-            // only three regions are available for single-account model. it's eu, us and br. 
+            // only three regions are available for single-account model. it's eu and us.
```
asein-sinch, line 135: Then there are only 2 regions, right?

src/Sinch/SinchClient.cs
```diff
@@ -170,5 +202,25 @@ private static string GetSmsRegion(SmsRegion smsRegion)
...
+            if (string.IsNullOrEmpty(appSecret))
+            {
+                throw new ArgumentNullException(nameof(appSecret), "The value should be present");
+            }
+            
+            var basicAuth = new BasicAuth(appKey, appSecret);
+            // TODO: implement application signed authentication, create IHttp just before the request with SignedRequestAuth
+            var http = new Http(basicAuth, _httpClient, _loggerFactory?.Create<Http>(), JsonNamingPolicy.CamelCase);
+            return new SinchVerificationClient(appKey, appSecret, 
+                new Uri("https://verification.api.sinch.com/"),
```
asein-sinch, line 222: Do you now if we have other URLs for test systems for instance? And if customers are using them during their development phase?

src/Sinch/Verification/Report/Request/FlashCallVerificationReportRequest.cs
```diff
@@ -0,0 +1,30 @@
+﻿namespace Sinch.Verification.Report
+{
+    public class FlashCallVerificationReportRequest : IVerifyReportRequest
+    {
+        /// <inheritdoc />
+        public override string Method { get; } = "flashCall";
```
asein-sinch, line 6: Is the value always be `flashCall` for a `FlashCallVerificationReportRequest`. I see it as an example in the spec

src/Sinch/Verification/Start/VerificationStartRequest.cs
asein-sinch, line 1: Is there a reason why you separate the `Start` and the `Report` packages for the `Verification` tag?
And also is there a reason why the `Report` package is split between Request and Response but the `Start` package contains everything at the root?

src/Sinch/Auth/OAuth.cs
asein-sinch, line 1: !! Not related to this PR
I've noticed that you are checking the `expiresIn` property from the authentication endpoint but from what @650elx explained me, we should not rely on this field. The idea of the refresh algorithm was to wait for the server to return a 401 status code with the http header `www-authenticate` containing the keyword `expired`.
Here is the reference to the conversation: https://sinch.slack.com/archives/C056G4AAMAS/p1696337843226489

src/Sinch/SinchClient.cs
Dovchik: I will have `internal` constructor for passing `baseUrl`, and have no intention to provide possibility to overwrite that in public api

src/Sinch/Verification/Report/Request/FlashCallVerificationReportRequest.cs
Dovchik: Yes, it will always be the same, see: https://developers.sinch.com/docs/verification/api-reference/verification/tag/Verifications/#tag/Verifications/operation/ReportVerificationByIdentity!c=200&path=0/method&t=response

src/Sinch/Verification/Start/VerificationStartRequest.cs
Dovchik: I try to do most narrow scoping, and then moving up common models. But with Request and Respons in report it's  oversight. I will do regroups in other PR

src/Sinch/Auth/OAuth.cs
Dovchik: For handling unauthorized, see [Http](https://github.com/sinch/sinch-sdk-dotnet/blob/f0f9b9cc2ade4569ac94f68a56cee797e6cd2047/src/Sinch/Core/Http.cs#L100) impelementation

JPPortier: I mean reference is about
`             if (_token is not null && _expiresIn.HasValue && DateTime.UtcNow < _expiresIn.Value && !force)`
and https://github.com/sinch/sinch-sdk-dotnet/blob/b21312ea6235def2b93c453ce943517f981955a6/src/Sinch/Auth/OAuth.cs#L91-L95
According to discussion pointed by @asein-sinch we only base invalidity onto server response and not computing ourselves the end of validity period

src/Sinch/SinchClient.cs
```diff
@@ -131,7 +162,7 @@ public class SinchClient : ISinch
         internal SinchClient(string projectId, Uri authUri, Uri numbersBaseAddress, Uri smsBaseAddress)
         {
             var http = new HttpClient();
-            Auth = new Auth.Auth(authUri, http);
+            Auth = new Auth.OAuth(authUri, http);
```
JPPortier, line 166: The call from this function contains comment `For E2E tests only. Here you can override base addresses.`
Ideally a dedicated function used for test only should be present at this level 
Official released product will contains dedicated to test only function but should not

src/Sinch/Verification/Start/IVerificationResponse.cs
```diff
@@ -1,7 +1,35 @@
...
+            JsonSerializerOptions options)
+        {
+            var elem = JsonElement.ParseValue(ref reader);
+            var descriptor = elem.EnumerateObject().FirstOrDefault(x => x.Name == "method");
+            return descriptor.Value.GetString() switch
+            {
+                "sms" => JsonSerializer.Deserialize<SmsResponse>(ref reader),
+                "callout" => JsonSerializer.Deserialize<PhoneCallResponse>(ref reader),
+                "flashCall" => JsonSerializer.Deserialize<FlashCallResponse>(ref reader),
+                "seamless" => JsonSerializer.Deserialize<DataResponse>(ref reader),
```
JPPortier: Ideally, these constant should be shared because of of used across different sources   (VerificationResponseBase.cs)

src/Sinch/SinchClient.cs
Dovchik: The function is not present in available public api as it's internal, and footprint in the final assembly is negligible. Not worth seating adjusting the build pipeline to remove the constructor.

src/Sinch/Verification/VerificationType.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Sinch.Verification
+{
+    internal static class VerificationType
+    {
+        public const string Sms = "sms";
+        public const string Seamless = "seamless";
+        public const string PhoneCall = "callout";
+        public const string FlashCall = "Feamles";
```
asein-sinch, line 8: What is `Feamles` ?

tests/Sinch.Tests/Verification/VerificationTests.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        private JsonSerializerOptions _jsonSerializerOptions;
+
+        [Fact]
+        public void DeserializeVerificationStartResponse()
+        {
+            var data = new
+            {
+                id = "1234567890",
+                method = "sms",
```
asein-sinch, line 26: Maybe you can use the `VerificationType` constants if it makes sense in the tests.

src/Sinch/SinchClient.cs
asein-sinch: What is "zt stack" ?
If there are only 2 regions, then you can update the comment accordingly. It's confusing otherwise.

src/Sinch/SinchClient.cs
asein-sinch: Ok, so we assume that customers will always use the production environment, right?

src/Sinch/Verification/Report/Request/FlashCallVerificationReportRequest.cs
asein-sinch: Ok, I understand now that you have derived the VerifyReportRequest class for all kinds of requests 👍

src/Sinch/SinchClient.cs
Dovchik: Need to create other pr for that about rerouting to supported hosting regions as EU and US

src/Sinch/Auth/OAuth.cs
Dovchik: At the end it doesn't matter as Http will care of invalidity.

tests/Sinch.Tests/Verification/VerificationTests.cs
Dovchik: That's anonymous object which usually represent raw data for my tests

src/Sinch/Auth/OAuth.cs
asein-sinch: From what I can see in Http.cs, only the status code will be checked, not the `www-authenticate` header value to verify that the unauthorized status comes from an expired token.

src/Sinch/SinchClient.cs
asein-sinch: Hum, I'm confused now between your 2 statements
 - "have no intention to provide possibility to overwrite that in public api"
 - "they can overwrite host"
 What should they do to target a different host then?

Dovchik: Overwrite host on OS level, e.g. https://www.makeuseof.com/tag/modify-manage-hosts-file-linux/
Or, if user fancy, he can try to call `internal` method. 
By public api I mean there is no `public` constructor for `SinchClient` which can allow overwrite of the `baseUrl`
Nevertheless, I think this discussion is already out of scope of this PR.

src/Sinch/Verification/Start/VerificationStartRequest.cs
```diff
@@ -0,0 +1,90 @@
...
+        ///     Verification by placing a PSTN call to the user's phone and playing an announcement,
+        ///     asking the user to press a particular digit to verify the phone number.
+        /// </summary>
+        public static readonly VerificationMethod Callout = new("callout");
+        
+        /// <summary>
+        ///     Data verification. Verification by accessing internal infrastructure of mobile carriers to verify
+        ///     if given verification attempt was originated from device with matching phone number.
+        /// </summary>
+        public static readonly VerificationMethod Seamless = new("seamless");
```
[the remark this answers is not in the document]
Dovchik: Welp, that's to API devs, id is really present in the response 🤷

src/Sinch/Auth/OAuth.cs
Dovchik: Yes, I will add `www-authenticate` check to handler

asein-sinch (approved):
LGTM. Thanks for taking the time to answer to my questions ☺️

PR #11, opened 2023-10-19
Feat/verification api verification status

Dovchik:
Add [Verification Status](https://developers.sinch.com/docs/verification/api-reference/verification/tag/Verification-status/) calls. Further tests with doppelganger.


src/Sinch/Verification/VerificationMethod.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Sinch.Core;
+
+namespace Sinch.Verification
+{
+    /// <summary>
+    ///     The method of the verification request.
+    /// </summary>
+    /// <param name="Value"></param>
+    public record VerificationMethod(string Value) : EnumRecord(Value)
```
asein-sinch, line 9: Maybe this record can be renamed to `VerificationStatusMethod` to match the documentation

asein-sinch, line 9: Btw, is there a reason why this one is a `record` whereas the `VerificationTypeInternal` (which could be renamed to `VerificationMethod`) is a `class` ?

Dovchik: Internal is for use of as constants in code. Record is specifically for this param: https://developers.sinch.com/docs/verification/api-reference/verification/tag/Verification-status/#tag/Verification-status/operation/VerificationStatusByIdentity!in=path&path=method&t=request

asein-sinch: Up to you, but as it is also described as `method` in the Verifications API, I was thinking about mentioning the VerificationStatus API name to clearly make the distinction.

asein-sinch (approved):

PR #12, opened 2023-10-20
Feat/tests for verification

Dovchik:
Tests will be green when doppleganger pr is merged

Covers:

- Verification start 

- Report Verification

- GetVerificationStatus

### DIFF
--- a/src/Sinch/SinchClient.cs
+++ b/src/Sinch/SinchClient.cs
@@ -91,10 +91,23 @@ namespace Sinch
 
     public class SinchClient : ISinch
     {
+        private const string VerificationApiUrl = "https://verification.api.sinch.com/";
+        private const string NumbersApiUrl = "https://numbers.api.sinch.com/";
+        private const string SmsApiUrlTemplate = "https://zt.{0}.sms.api.sinch.com";
+        private const string ConversationApiUrlTemplate = "https://{0}.conversation.api.sinch.com/";
+
         private readonly LoggerFactory _loggerFactory;
         private readonly HttpClient _httpClient;
         private readonly Uri _verificationBaseAddress = null;
 
+        /// <summary>
+        ///     Initialize a new <see cref="SinchClient"/>
+        /// </summary>
+        /// <param name="keyId">Your Sinch Account key id.</param>
+        /// <param name="keySecret">Your Sinch Account key secret.</param>
+        /// <param name="projectId">Your project id.</param>
+        /// <param name="options">Optional. See: <see cref="SinchOptions"/></param>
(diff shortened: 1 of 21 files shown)


src/Sinch/SinchClient.cs
```diff
@@ -210,16 +214,17 @@ public ISinchVerificationClient Verification(string appKey, string appSecret)
...
             }
-            
+
             var basicAuth = new BasicAuth(appKey, appSecret);
             // TODO: implement application signed authentication, create IHttp just before the request with SignedRequestAuth
             var http = new Http(basicAuth, _httpClient, _loggerFactory?.Create<Http>(), JsonNamingPolicy.CamelCase);
-            return new SinchVerificationClient(appKey, appSecret, 
-                new Uri("https://verification.api.sinch.com/"), 
+            return new SinchVerificationClient(appKey, appSecret,
+                _verificationBaseAddress ?? new Uri("https://verification.api.sinch.com/"),
```
JPPortier, line 227: Just a quick note.
Personally, having all this kind of constants (URLs for endpoints) centralized in a constant section within the file (or a constant file if this is having sense) facilitate reading sources. 
And if looking for external dependencies it is not "hidden" across files/lines but clearly identified and easier to modify

Dovchik: Yeah, easier in case if variable is used more than one time, but when it's the only place then it's makes sense to keep to place of use as near as possible including inlining. But I'll introduce those constants as it makes sense to have them for fast understanding what are these about

src/Sinch/SinchClient.cs
```diff
@@ -123,34 +137,36 @@ public class SinchClient : ISinch
...
+            Auth = auth;
+
+            logger?.LogInformation("SinchClient initialized.");
+        }
 
+        private static Uri GetSmsBaseAddress(SmsRegion smsRegion)
+        {
             // only three regions are available for single-account model. it's eu and us. 
             // So, we should map other regions provided in docs to nearest server.
             // See: https://developers.sinch.com/docs/sms/api-reference/#base-url
```
JPPortier, line 163: This comment is duplicated (but different) with the one describing the `GetSmsRegion` function.
May be removing this comment is fine;
Note for further discussion/ticket regarding to the `getSmsRegion` implementation: do we still want to automatically and silently move to another region ?

Dovchik: I already have a task for that which I will plan to take in the next sprint

JPPortier: (just reminder about duplicated but different comment regarding `GetSmSregion` call vs implementation)

PR #22, opened 2023-11-21
Feat/introduce voice api and callouts

Dovchik:


tests/Sinch.Tests/Voice/VoiceClientTests.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Sinch.Tests.Voice
+{
+    public class VoiceClientTests
+    {
+        [Fact]
+        public void InitVoiceWithGlobalRegion()
+        {
+            var client = new SinchClient("id", "secret", "id");
+            var voiceClient = client.Voice("key", "secret");
```
JPPortier, line 15: This is highlighting a pending question about unified authentication way for SDK in front of multiple APIs
Here, the voiceClient usage with dedicated key/secret (vs the ones used for unified SinchClient) is highlighting differences and not an unified way

Dovchik: I'm confused. The clients are different. Can you elaborate what is expected, and what should we do to procede with review?

JPPortier (approved):

PR #25, opened 2023-11-30
Feat/drop net5

Dovchik:
net5 is out of support, so dropping this https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core


650elx (approved):
LGTM

PR #30, opened 2024-01-04
Voice: add callbacks

Dovchik:
Add classes for use with [Voice Callbacks](https://developers.sinch.com/docs/voice/api-reference/voice/tag/Callbacks/)
Add JsonPropertyNames for related callback classes to override json naming convention of any used json serializer


650elx (approved):
🆗

PR #36, opened 2024-01-29
Feat/conversation conversations

Dovchik:
Covers Conversations/Conversation tag of api.
Also, includes rework of Conversation/Messages models to match server json schema


examples/Console/AccessingPolymorphicType.cs
```diff
@@ -0,0 +1,27 @@
...
+using Sinch.Verification.Start.Response;
+
+namespace Examples
+{
+    public class AccessingPolymorphicType
+    {
+        public static void Example()
+        {
+            var sinchClient = new SinchClient("KEY_ID", "KEY_SECRET", "PROJECT_ID");
+            var verificationStart = sinchClient.Verification("APP_KEY", "APP_SECRET").Verification
```
asein-sinch, line 12: You may want to rename this variable `verificationStartResponse` to be consistent with the other variables below or `verificationStartResult` to be consistent with the function call.

src/Sinch/Conversation/Conversations/Conversations.cs
```diff
@@ -0,0 +1,233 @@
...
+            return _http.Send<Conversation, Conversation>(builder.Uri, HttpMethod.Patch, conversation,
+                cancellationToken);
+        }
+
+        /// <inheritdoc />
+        public Task InjectMessage(InjectMessageRequest injectMessageRequest,
+            CancellationToken cancellationToken = default)
+        {
+            if (string.IsNullOrEmpty(injectMessageRequest.ConversationId))
+                throw new NullReferenceException(
```
asein-sinch, line 222: Sometimes you are throwing a `NullReferenceException` and other times it's a `ArgumentNullException`

asein-sinch (approved):
Approved with some small comments to look at

src/Sinch/Conversation/Conversations/Conversations.cs
```diff
@@ -0,0 +1,233 @@
...
+                throw new NullReferenceException($"{nameof(conversation.Id)} should have a value");
+
+            var builder = new UriBuilder(new Uri(_baseAddress,
+                $"v1/projects/{_projectId}/conversations/{conversation.Id}"));
+
+            var queryString = HttpUtility.ParseQueryString(string.Empty);
+            var propMask = conversation.GetPropertiesMask();
+            if (!string.IsNullOrEmpty(propMask)) queryString.Add("update_mask", propMask);
+
+            if (metadataUpdateStrategy is not null)
```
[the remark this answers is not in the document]
Dovchik: [Isn't it optional](https://developers.sinch.com/docs/conversation/api-reference/conversation/tag/Conversation/#tag/Conversation/operation/Conversation_UpdateConversation!in=query&path=metadata_update_strategy&t=request) and going to be defaulted by the server?

src/Sinch/Conversation/Conversations/Conversations.cs
Dovchik: True - it's intended, `ArgumentNullException` is obviously for argument of a function, null ref for is for a other props of an object in this case.

asein-sinch: So in this case, you may also want to check the argument is not null before checking its properties in order to avoid a NPE

PR #37, opened 2024-01-31
Feat/voice dedicated start verification functions

Dovchik:
Also, fix deserialization of IVerificationStartResponse

### DIFF
--- a/src/Sinch/Verification/SinchVerification.cs
+++ b/src/Sinch/Verification/SinchVerification.cs
@@ -20,7 +20,43 @@ namespace Sinch.Verification
         /// <param name="request"></param>
         /// <param name="cancellationToken"></param>
         /// <returns></returns>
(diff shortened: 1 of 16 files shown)


src/Sinch/Verification/SinchVerification.cs
```diff
@@ -20,7 +21,56 @@ public interface ISinchVerification
         /// <param name="request"></param>
         /// <param name="cancellationToken"></param>
         /// <returns></returns>
-        Task<IVerificationStartResponse> Start(VerificationStartRequest request,
+        Task<IStartVerificationResponse> Start(StartVerificationRequest request,
```
asein-sinch, line 24: Do we need to keep this method if all the 4 possibilities are covered below?

Dovchik: yeah, makes sense - now this method is an overhead

asein-sinch (approved):

PR #46, opened 2024-03-26
Fix/conversation models

Dovchik:
To reflect update of API conversation models is updated.


src/Sinch/Conversation/Messages/Message/CardMessage.cs
```diff
@@ -59,7 +59,5 @@ public override string ToString()
     public class MediaCarouselMessage
     {
         public Uri Url { get; set; }
-
-        public string Caption { get; set; }
```
JPPortier, line 63: Great.
We definitively need to be able to identify such breaking change proactively

JPPortier (approved):

PR #47, opened 2024-03-28
Feat/support service plan

Dovchik:
Add support for Service plan id of sms api. Which can be set via SinchOptions

### DIFF
--- a/examples/Console/UseServicePlanIdForSms.cs
+++ b/examples/Console/UseServicePlanIdForSms.cs
@@ -1,0 +1,26 @@
+﻿using Sinch;
+using Sinch.SMS;
+using Sinch.SMS.Batches.Send;
+
+namespace Examples
+{
+    public class UseServicePlanIdForSms
+    {
+        public void Example()
+        {
+            var sinchClient = new SinchClient(default, default, default, options =>
+            {
+                options.UseServicePlanIdWithSms(Environment.GetEnvironmentVariable("SINCH_SERVICE_PLAN_ID"),
+                    SmsServicePlanIdHostingRegion.Ca, Environment.GetEnvironmentVariable("SINCH_API_TOKEN"));
+            });
+            sinchClient.Sms.Batches.Send(new SendTextBatchRequest()
+            {
+                Body = "Hello, World!",
+                To = new List<string>()
+                {
+                    "+480000000"
+                }
+            });
+        }
+    }
+}
--- a/examples/Console/UseServicePlanIdForSms.cs
+++ b/examples/Console/UseServicePlanIdForSms.cs
@@ -8,11 +8,12 @@ namespace Examples
     {
         public void Example()
         {
-            var sinchClient = new SinchClient(default, default, default, options =>
-            {
-                options.UseServicePlanIdWithSms(Environment.GetEnvironmentVariable("SINCH_SERVICE_PLAN_ID"),
-                    SmsServicePlanIdHostingRegion.Ca, Environment.GetEnvironmentVariable("SINCH_API_TOKEN"));
-            });
+            var sinchClient = new SinchClient(default, default, default,
+                options =>
+                {
+                    options.UseServicePlanIdWithSms(Environment.GetEnvironmentVariable("SINCH_SERVICE_PLAN_ID"),
+                        Environment.GetEnvironmentVariable("SINCH_API_TOKEN"), SmsServicePlanIdHostingRegion.Ca);
+                });
             sinchClient.Sms.Batches.Send(new SendTextBatchRequest()
             {
                 Body = "Hello, World!",
--- a/src/Sinch/SMS/SmsServicePlanIdHostingRegion.cs
+++ b/src/Sinch/SMS/SmsServicePlanIdHostingRegion.cs
@@ -3,7 +3,7 @@
     /// <summary>
     ///     The following regions can be set to be used in SDK as a hosting region when using `service_plan_id` option
     /// </summary>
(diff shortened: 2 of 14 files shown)


src/Sinch/SMS/SmsServicePlanIdHostingRegion.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Sinch.SMS
```
JPPortier, line 1: Not sure this will be an added value from SDK user's perspective : defining a dedicated type for supported region in case of service plan id usage
This mean if for test or any other reason user switch from `projectId` <-> `servicePlanId` usage the region as also to be changed. Not because wanting to changed to the region but because type is different.
e.g.: Still addressing `us` region in both cases but have to change `SmsHostingRegion.Us` <-> `SmsServicePlanIdHostingRegion.Us`
Even if supported regions are not equals, a dynamic validation against unexpected values should be enough.
And even in this case: it should be a warning and not an error. So in case of server going forward to support new region, SDK's user do not need to move to a new (but not yet ready release) SDK updated version

src/Sinch/SinchClient.cs
```diff
@@ -353,5 +291,63 @@ public ISinchAuth Auth
...
+            }
+
+            _logger?.LogInformation("Initializing SMS client with {project_id} in {region}", _projectId,
+                optionsObj.SmsHostingRegion.Value);
+            return new SmsClient(_projectId, GetSmsBaseAddress(optionsObj.SmsHostingRegion, _apiUrlOverrides?.SmsUrl),
+                _loggerFactory,
+                httpSnakeCase);
+        }
+
+        private static Uri GetSmsBaseAddress(SmsServicePlanIdHostingRegion smsServicePlanIdHostingRegion,
```
JPPortier, line 333: Related to previous comment about different types for same region usage: method signature based onto region type is used to trigger `servicePlanId` vs `projectId` URL.
This should be related to `servicePlanId` usage not onto region type usage

src/Sinch/SinchOptions.cs
```diff
@@ -38,6 +39,54 @@ public sealed class SinchOptions
...
+
+        /// <summary>
+        ///     Use SMS API with `service plan id` and compatible region.
+        ///     `service_plan_id` will be used in place of `project_id`
+        /// </summary>
+        /// <param name="servicePlanId">Your service plan id</param>
+        /// <param name="hostingRegion">Region to use.</param>
+        /// <param name="apiToken"></param>
+        /// <exception cref="ArgumentNullException">throws if service plan id or region is null or an empty string</exception>
+        public void UseServicePlanIdWithSms(string servicePlanId, SmsServicePlanIdHostingRegion hostingRegion,
```
asein-sinch, line 54: The order of parameters seems strange: the documentation says "Default server location is USA" so I would expect the hosting region parameter to be optional.
On the other hand, the apiToken is mandatory and I would expect to see it next to the servicePlanId

examples/Console/UseServicePlanIdForSms.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Examples
+{
+    public class UseServicePlanIdForSms
+    {
+        public void Example()
+        {
+            var sinchClient = new SinchClient(default, default, default, options =>
+            {
+                options.UseServicePlanIdWithSms(Environment.GetEnvironmentVariable("SINCH_SERVICE_PLAN_ID"),
```
asein-sinch, line 13: It looks like the servicePlanId is a second class citizen for SMS: the user still have to go to the SinchClient initialization with the 3 parameters and then define a 4th parameter to provide the login informations

src/Sinch/SMS/SmsServicePlanIdHostingRegion.cs
Dovchik: I'll keep dedicated type for regions as sets of supported are different. For any 'I don't want checks' users it's a record so `new SmsServicePlainIdHostingRegion('cn')` should work.

src/Sinch/SinchClient.cs
Dovchik: I'm renaming the method to better reflect underlying functionality

src/Sinch/SinchOptions.cs
Dovchik: Makes sense, I'm updating the method signature

Dovchik: Also, SmsServicePlainIdHosting regions cannot be made optional due to language limitations

examples/Console/UseServicePlanIdForSms.cs
Dovchik: Yes, it is, I'm keeping it that way.

src/Sinch/SMS/SmsServicePlanIdHostingRegion.cs
JPPortier: From the SDK user's point of view, a region is just a region: the EU, the USA, ... are no different targets when approached by service plan or project.
What is the added value of complexity for SDK's user?

Dovchik: The supported set of regions  _are_ different, and different enums reflects just that.

asein-sinch: This has to be discussed within the team. Item added in the parking lot

asein-sinch (approved):
As discussed during the standup meeting, we'll address the regions handling after it's discussed in details within the team

PR #48, opened 2024-04-03
Feat/fax api faxes

Dovchik:
Introduce Fax Api support 

This PR brings in FaxClient interface and implementation of Faxes tag. 

This is a PR sink for other fax related PRs as well.

Will be merged only when other child PRs regarding other tags will be merge into current.


src/Sinch/Core/Http.cs
```diff
@@ -80,23 +86,87 @@ internal class Http : IHttp
...
+
+                var type = value.GetType();
+                if (type == typeof(List<string>))
+                {
+                    var asString = string.Join(',', (value as List<string>)!);
+                    content.Add(new StringContent(asString), prop.Name);
+                }
+                else if (type == typeof(Dictionary<string, string>))
+                {
+                    foreach (var (key, val) in (value as Dictionary<string, string>)!)
```
asein-sinch, line 112: I would have expected to see a recursion for such type. How do you handle nested structures?

src/Sinch/Core/Http.cs
```diff
@@ -80,23 +86,87 @@ internal class Http : IHttp
                 $"sinch-sdk/{sdkVersion} (csharp/{RuntimeInformation.FrameworkDescription};;)";
         }
 
+        public Task<TResponse> SendMultipart<TRequest, TResponse>(Uri uri, TRequest request, Stream stream,
+            string fileName, CancellationToken cancellationToken = default)
+        {
+            var content = new MultipartFormDataContent();
```
asein-sinch, line 92: I think this part deserves to be put in a dedicated function such as `ConvertToMultipartFormDataContent` and be unit tested

src/Sinch/Fax/Faxes/Utils.cs
```diff
@@ -0,0 +1,431 @@
...
+using System.Collections.Generic;
+
+namespace Sinch.Fax.Faxes
+{
+    internal class FileExtensionContentTypeProvider
+    {
+        private IDictionary<string, string> Mappings { get; set; }
+
+
+        internal FileExtensionContentTypeProvider()
```
asein-sinch, line 11: Where does this list come from? I think it can be trimmed down to the file types supported by the API

src/Sinch/Core/StreamExtensions.cs
```diff
@@ -0,0 +1,23 @@
+using System;
+using System.IO;
+
+namespace Sinch.Core
+{
+    internal static class StreamExtensions
+    {
+        public static string ConvertToBase64(this Stream stream)
```
asein-sinch, line 8: Where are you using this method? I can't find any occurence in the diff displayed by github

src/Sinch/Core/Http.cs
```diff
@@ -186,5 +283,22 @@ internal class Http : IHttp
...
+        {
+            HttpContent? httpContent =
+                request == null ? null : JsonContent.Create(request, options: _jsonSerializerOptions);
+
+
+            return await SendHttpContent<TResponse>(uri: uri, httpMethod: httpMethod, httpContent,
+                cancellationToken: cancellationToken);
+        }
+
+        public Task<TResponse> SendMultipart<TRequest, TResponse>(Uri uri, HttpMethod httpMethod, TRequest request,
```
asein-sinch, line 298: What is the reason to declare this method?

examples/Console/Fax/DownloadFax.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Examples.Fax
+{
+    public class DownloadFax
+    {
+        public static async Task Example()
+        {
+            var sinchClient = new SinchClient("PROJECT_ID", "KEY_ID", "KEY_SECRET");
+            await using var responseStream = await sinchClient.Fax.Faxes.DownloadContent("FAX_ID");
+            await using var fileStream = new FileStream("C:\\Downloads\\fax.pdf", FileMode.Create, FileAccess.Write);
```
asein-sinch, line 11: If the directory doesn't exist, will it create it?

asein-sinch, line 11: It can be worth to show how to read the file name from the response rather than hardcoding it

asein-sinch (commented):
Nice work! I've put some comments that also apply to the implementation of this API in the Node.js SDK 🙂

Dovchik (commented):
Thanks for the review Antoine! Here you will find adjustment to mocks https://github.com/sinch/doppelganger/pull/25

examples/Console/Fax/DownloadFax.cs
asein-sinch: Yes, it's worth to illustrate how to read the file name

src/Sinch/Fax/Faxes/Fax.cs
```diff
@@ -0,0 +1,236 @@
...
+        ///     A phone number in [E.164](https://community.sinch.com/t5/Glossary/E-164/ta-p/7537) format, including the leading &#39;+&#39;.
+        /// </summary>
+        [JsonPropertyName("from")]
+        public string? From { get; set; }
+
+        /// <summary>
+        ///     A phone number in [E.164](https://community.sinch.com/t5/Glossary/E-164/ta-p/7537) format, including the leading &#39;+&#39;.
+        /// </summary>
+        [JsonPropertyName("to")]
+        public string? To { get; set; }
```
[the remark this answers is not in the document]
asein-sinch: Indeed, this is what is written in the documentation but the specification is not reflecting this information 😞

src/Sinch/Fax/Faxes/FaxesClient.cs
```diff
@@ -0,0 +1,180 @@
...
+        public Task<Stream> DownloadContent(string id, CancellationToken cancellationToken = default)
+        {
+            if (string.IsNullOrEmpty(id))
+            {
+                throw new ArgumentNullException(nameof(id), "Fax id should have a value.");
+            }
+
+            _loggerAdapter?.LogInformation("Downloading the content of the fax with {id}", id);
+            var uriBuilder = new UriBuilder(_uri);
+            uriBuilder.Path += $"/{id}/file.pdf"; // only pdf is supported for now
```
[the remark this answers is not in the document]
asein-sinch: Instead of a PDF, the server may generate TIFF for instance. We'll see whether it evolves some day.

src/Sinch/Fax/Faxes/ListFaxesRequest.cs
```diff
@@ -0,0 +1,136 @@
...
+            var queryString = new List<KeyValuePair<string, string>>();
+
+            if (CreateTime.HasValue) // mutually exclusive 
+            {
+                queryString.Add(new("createTime",
+                    StringUtils.ToIso8601(CreateTime.Value)));
+            }
+            else
+            {
+                if (CreateTimeAfter.HasValue)
```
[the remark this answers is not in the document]
asein-sinch: I'm sorry, but I don't understand your comment. Where is the code that creates the inclusive comparator `>=`.
And for the second part of my comment, I'm not sure you have answered to that.

src/Sinch/Core/Utils.cs
```diff
@@ -144,4 +189,10 @@ public static void ThrowNullDeserialization(Type type)
             throw new InvalidOperationException($"{type.Name} deserialization result is null");
         }
     }
+
+    internal enum PageStart
+    {
+        Zero = 0,
+        First = 1,
```
asein-sinch, line 196: The naming is misleading as we are looking for the first page index in all cases. I think "One" would be be more understandable. Or `IndexZero` and `IndexOne`

src/Sinch/Fax/Faxes/SendFaxRequest.cs
```diff
@@ -0,0 +1,182 @@
...
+        /// </summary>
+        /// <param name="filePath"></param>
+        public SendFaxRequest(string filePath)
+        {
+            FileContent = File.OpenRead(filePath);
+            FileName = Path.GetFileName(filePath);
+        }
+
+        /// <summary>
+        ///     A phone number in [E.164](https://community.sinch.com/t5/Glossary/E-164/ta-p/7537) format, including the leading &#39;+&#39;.
```
asein-sinch, line 54: As you have enforced the usage of a list, this comment needs to be updated

src/Sinch/Fax/Faxes/FaxesClient.cs
```diff
@@ -0,0 +1,196 @@
...
+        ///     Create and send a fax.<br/><br/>
+        ///     Fax content may be supplied via one or more files or URLs of supported filetypes.<br/><br/>
+        ///     If you supply a callbackUrl the callback will be sent as multipart/form-data with the content
+        ///     of the fax as an attachment to the body, unless you specify callbackUrlContentType as application/json.
+        ///     If you specify more than one receiver, list will have more then one faxes.
+        /// </summary>
+        /// <param name="request"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public Task<List<Fax>> Send(SendFaxRequest request, CancellationToken cancellationToken = default);
```
asein-sinch, line 31: I see you have decided to enforce the usage of an array of faxes for the response. For the request part, this is similar to what is proposed for SMS batches, but for the response, it may be an additional burden for the SDK user to have to manage a list of faxes when there is a single recipient.

Dovchik: Okay, I can play around it

src/Sinch/Fax/Faxes/ListFaxesRequest.cs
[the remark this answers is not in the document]
Dovchik: Regarding inclusiveness, `ToQueryString` below will convert `createTime>` + `dateTime` to `createTime>=dateTime`

> And the documentation also specifies that the date can shorten up to the month (and I tested, even up to the year)

Yes, but there is no neat way to represent that with `DateTime` or `DateOnly` structs, and no need, with them user will be enough to query any range in handy manner.

Dovchik:
Added an override for send fax https://github.com/sinch/doppelganger/pull/26

examples/Console/Fax/DownloadFax.cs
asein-sinch: This is what I'm doing in Node.js: the return type is an object that contains the fileName extracted from the content-disposition header and the buffer. Can you elaborate why you consider this structure not that clean?
And it's true that the filename returned by the server is the faxId, but hardcoding it in the example seems a hack to compensate for the loss of information from the header.

Dovchik: I'll just implement that

src/Sinch/Core/Http.cs
```diff
@@ -263,13 +263,19 @@ bool HasNonNullValue(PropertyInfo x)
                 // NOTE: there wil probably be other files supported in the future
                 if (result.IsPdf())
                 {
-                    if (typeof(TResponse) != typeof(Stream))
+                    if (typeof(TResponse) != typeof(ContentResult))
```
asein-sinch, line 266: I don't understand how the type of response would be different from the previous state where it was a Stream. Unless I'm mistaken, the `ContentResult` object will be created on lines 274-278 below, right?

Dovchik: Yes, ContentResult will be created below. The response as read as a raw stream, but now it's contained in `ContentResult` class. It's a safe measure when SDK dev will try to use another class which is not compatible with raw stream content.

asein-sinch (approved):
Well done. Regarding the PR organization, maybe you can create another branch from main where all the sub-features for Fax (this PR being one of the sub-features) will be merged. Otherwise, I fear it will be a mess if you need to come back to the Faxes part once Email will be merged

PR #50, opened 2024-04-08
fix: add specific object into list message

Dovchik:
Update to follow model: https://developers.sinch.com/docs/conversation/api-reference/conversation/tag/Messages/#tag/Messages/operation/Messages_SendMessage!path=message/7/list_message/sections/items/item/0/choice&t=request


src/Sinch/Conversation/Messages/Message/ListMessage.cs
```diff
@@ -95,6 +97,31 @@ public interface IListItem
     {
     }
 
+    public class ListItem
```
JPPortier, line 100: Not related to this dedicated PR, but there is a design point here (existing code is already having this e. g. https://github.com/sinch/sinch-sdk-dotnet/blob/main/src/Sinch/Conversation/Messages/Message/AppMessage.cs#L14)
This is the direct translation/representation of the oneOf from OAS
From a SDK user point of view, it is possible to fill both fields and send them to server, right ?
We should avoid this at SDK level and provide a way to fill a single field with the dedicated type
This subject is planning for next meeting where we could discuss more about it

src/Sinch/Conversation/Messages/Message/ListMessage.cs
```diff
@@ -72,7 +74,7 @@ public sealed class ListSection
         /// <summary>
         ///     Gets or Sets Items
         /// </summary>
-        public List<IListItem> Items { get; set; }
+        public List<ListItem> Items { get; set; }
```
JPPortier, line 77: Waiting feedback from dedicated team but seems all items should be of same type.
So this could imply to have something like `List<ListItem<T>>` where `T` could be `ListItemChoice` or `ListItemProduct`
(or even at `ListSection` level)

src/Sinch/Conversation/Messages/Message/ListMessage.cs
Dovchik: No, you can only send `ListItemChoise` or `ListItemProduct`, you cannot send both simultaneously.

This is achieved with private set and constructor which takes the specific type as a param. 

public default constructor is there, BUT `Obsolete` attribute with `error: True` use will result in a compiler error.

JPPortier: Ok. Got it
Thank for C# growth/knowledge 👌

src/Sinch/Conversation/Messages/Message/ListMessage.cs
JPPortier: Just received confirmation: All sections have to be of the same type in a ListMessage

Dovchik: But, unfortunately, I don't see how can I simplify the structure anyway due to specifics of JSON structure of that response. 😕  So the way the things are middle-ground between specification and code, which I would like to merge as is anyway.

JPPortier: What about using same design as the one used by SMS domain ? 
It could offer benefits:
1. using native OO C# language
2. same design across domains within the SDK (currently SMS is using  base class + inheritance for oneOf/allOf implementation)
3. once OO classes, this will enable to have ListItem<T> and ListSection<T> to restrict whole list items to be of same type

Out of this dedicated PR but highlighted by this one: point 1. and 2. are about whole Conversation domain. So design from one domain to another is not offering a consistant experience to SDK users

src/Sinch/Conversation/Messages/Message/ListMessage.cs
```diff
@@ -95,6 +97,16 @@ public interface IListItem
...
     }
 
+    public class ListItemChoiceWrapper : IListItem
+    {
+        public ListItemChoice Choice { get; set; }
+    }
+
+    public class ListItemProductWrapper : IListItem
+    {
+        public ListItemProduct Product { get; set; }
```
JPPortier, line 107: Not dedicated to this PR, but last specs are now using `ProductItem` schema for `ProductResponseMessage` schema too.
So the type named `ListItemProduct` is not longer pertinent in the `ProductResponseMessage` context when it will be implemented

src/Sinch/Conversation/Messages/Message/ListMessage.cs
```diff
@@ -95,6 +97,16 @@ public interface IListItem
     {
     }
 
+    public class ListItemChoiceWrapper : IListItem
```
JPPortier, line 100: Do you think there is a way to hide the wrapper required onto JSON serialization only ?

The before/after samples from this PR (https://github.com/sinch/sinch-sdk-dotnet/pull/50/files#diff-e1497a7d723d2a4ac1227f7ef48afc83fe25163f5b2f07c76222feb1f7e2487fR48) highlight complexity of network protocol without added value for SDK's user.

tests/Sinch.Tests/Conversation/MessagesTests.cs
```diff
@@ -45,27 +45,33 @@ public async Task GetMessage()
                         Title = "sec1",
                         Items = new List<IListItem>()
                         {
-                            new ListItemChoice()
+                            new ListItemChoiceWrapper
                             {
-                                Title = "title",
-                                Description = "desc",
-                                Media = new MediaMessage()
+                                Choice = new ListItemChoice()
```
JPPortier, line 50: Could be nice to not having to create wrapper + item and having a single type to be filled

Dovchik: flattened

JPPortier (approved):

PR #53, opened 2024-04-12
refactor(voice): refs?

Dovchik:
For models any non-required/optional field is nullable

Add EmptyResponse class for empty responses from the API, refactor api calls to use it


src/Sinch/Core/Http.cs
```diff
@@ -154,13 +161,27 @@ internal class Http : IHttp
...
+                           ?? throw new InvalidOperationException(
+                               $"{typeof(TResponse).Name} is null");
+
+                // if empty response is expected, any non related response is dropped
+                if (typeof(TResponse) == typeof(EmptyResponse))
+                {
+                    // if not empty content, check what is there for debug purposes.
+                    if (result.Content.GetType().Name != "EmptyContent")
+                    {
+                        _logger?.LogDebug("Response is not json, but {content}",
```
asein-sinch, line 173: I'm not sure about the log here: shouldn't it be something like "Response is not empty" ?

src/Sinch/Voice/Callouts/Callout/CustomCalloutRequest.cs
```diff
@@ -62,6 +65,6 @@ public class CustomCalloutRequest
         ///     This could result in further SVAML or some other application logic function.
         ///     <example>"https://your-application-server-host/application"</example>
         /// </summary>
-        public string Pie { get; set; }
+        public string? Pie { get; set; }
```
asein-sinch, line 68: Similarly to `ace` and `ice`, the `pie` property can also be a string representation of a JsonObject

src/Sinch/Voice/Callouts/Callout/TextToSpeechCalloutRequest.cs
```diff
@@ -76,7 +76,7 @@ public class TtsCalloutRequest
         ///     Example:
         ///     <c>#tts[Hello from Sinch];#ssml[&lt;speak&gt;&lt;break time=\"250ms\"/&gt;Have a great day!&lt;/speak&gt;]</c>
         /// </summary>
-        public string Promts { get; set; }
+        public string? Promts { get; set; }
```
asein-sinch, line 79: Typo

src/Sinch/Voice/Calls/Call.cs
```diff
@@ -60,19 +61,19 @@ public sealed class Call
         /// <summary>
         ///     An object that can be used to pass custom information related to the call.
         /// </summary>
-        public object Custom { get; set; }
+        public JsonObject? Custom { get; set; }
```
asein-sinch, line 64: Why this change? The documentation mentions a `string`: https://developers.sinch.com/docs/voice/api-reference/voice/tag/Calls/#tag/Calls/operation/Calling_GetCallResult!c=200&path=custom&t=response

asein-sinch (commented):
Review done. I checked the parts related to Conversation and SMS too, even if the PR mentioned voice only

src/Sinch/Voice/Callouts/Callout/CustomCalloutRequest.cs
Dovchik: I'll update it in separate PR

src/Sinch/Voice/Callouts/Callout/TextToSpeechCalloutRequest.cs
Dovchik: That will be in separate as well

src/Sinch/Voice/Callouts/Callout/CustomCalloutRequest.cs
asein-sinch: I don't understand what qualifies a change to be in a PR or not. If you don't plan to update this property, are you going to revert what you did for `Ace` and `Ice` just above?

src/Sinch/Voice/Callouts/Callout/TextToSpeechCalloutRequest.cs
asein-sinch: In this case, please create a ticket and communicate the reference to the reviewers

src/Sinch/Voice/Calls/Call.cs
asein-sinch: In this case, maybe you can leave it as it was and fix the type only once in a separate PR?

asein-sinch (approved):

PR #56, opened 2024-04-22
feat(example): add handling of incoming call event

Dovchik:


examples/WebApi/Controllers/HandleIncomingIceEventController.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        private readonly ISinchClient _sinchClient;
+
+        public HandleIncomingIceEventController(ISinchClient sinchClient)
+        {
+            _sinchClient = sinchClient;
+        }
+
+        [HttpPost]
+        [Route("ice-event")]
```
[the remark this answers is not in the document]
Dovchik: The server will return 400 Bad Request for other events. But probably it makes sense to make endpoint generic with matching for specific event...

Dovchik:
Added refactor for events to have generic interface which can be used for pattern matching

asein-sinch (approved):

examples/WebApi/Controllers/HandleIncomingIceEventController.cs
```diff
@@ -0,0 +1,55 @@
...
+        public HandleIncomingIceEventController(ISinchClient sinchClient)
+        {
+            _sinchClient = sinchClient;
+        }
+
+        [HttpPost]
+        [Route("event")]
+        public IActionResult HandleEvent([FromBody] IVoiceEvent incomingEvent)
+        {
+            switch (incomingEvent)
```
JPPortier, line 24: Ideally Voice should provide a `ValidateAuthenticationHeader` function (like Conversation domains: https://github.com/sinch/sinch-sdk-dotnet/blob/main/src/Sinch/Conversation/Webhooks/Webhooks.cs#L72)

Calling `ValidateAuthenticationHeader` here and because this sample is a tutorial, it will help to highlight endpoint is having to be "protected" against unsolicited requests

examples/WebApi/Controllers/HandleIncomingIceEventController.cs
```diff
@@ -0,0 +1,55 @@
...
+                        Instructions = new List<IInstruction>()
+                        {
+                            new Say()
+                            {
+                                Text = "Thank you for calling Sinch! This call will now end.",
+                                Locale = "en-US"
+                            }
+                        }
+                    };
+                    return Ok(response);
```
JPPortier, line 43: _Side note: like the OK/BadRequest; need to think about having same "helper" for Java SDK to hide/help the response serialization requirements_

src/Sinch/Voice/Hooks/AnsweredCallEvent.cs
```diff
@@ -64,5 +63,7 @@ public class AnsweredCallEvent
         /// </summary>
         [JsonPropertyName("amd")]
         public Amd? Amd { get; set; }
+
+        public EventType? EventType { get; }
```
JPPortier, line 67: Seems this field is not coming present from specs. Could be removed?

examples/WebApi/Controllers/HandleIncomingIceEventController.cs
Dovchik: Makes sense, blocked by #62

src/Sinch/Voice/Hooks/AnsweredCallEvent.cs
JPPortier: ⚠️ This PR was merged, but this field is still defined

Dovchik: Okay, re checked  - seems like false alarm @JPPortier

PR #61, opened 2024-04-24
fix(voice): update models

Dovchik:
Updates some de synchronized models. 

For tests https://github.com/sinch/doppelganger/pull/18 should be merged before

### DIFF
--- a/src/Sinch/Voice/Callouts/Callout/CustomCalloutRequest.cs
+++ b/src/Sinch/Voice/Callouts/Callout/CustomCalloutRequest.cs
@@ -65,6 +65,7 @@ namespace Sinch.Voice.Callouts.Callout
         ///     This could result in further SVAML or some other application logic function.
         ///     <example>"https://your-application-server-host/application"</example>
         /// </summary>
-        public string? Pie { get; set; }
+        [JsonConverter(typeof(JsonObjectAsStringJsonConverter))]
+        public JsonObject? Pie { get; set; }
     }
 }
--- a/src/Sinch/Voice/Callouts/Callout/CustomCalloutRequest.cs
+++ b/src/Sinch/Voice/Callouts/Callout/CustomCalloutRequest.cs
@@ -65,7 +65,7 @@ namespace Sinch.Voice.Callouts.Callout
         ///     This could result in further SVAML or some other application logic function.
         ///     <example>"https://your-application-server-host/application"</example>
         /// </summary>
-        [JsonConverter(typeof(JsonObjectAsStringJsonConverter))]
-        public JsonObject? Pie { get; set; }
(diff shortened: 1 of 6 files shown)


src/Sinch/Voice/Callouts/Callout/CustomCalloutRequest.cs
```diff
@@ -65,6 +65,7 @@ public class CustomCalloutRequest
         ///     This could result in further SVAML or some other application logic function.
         ///     <example>"https://your-application-server-host/application"</example>
         /// </summary>
-        public string? Pie { get; set; }
+        [JsonConverter(typeof(JsonObjectAsStringJsonConverter))]
+        public JsonObject? Pie { get; set; }
```
asein-sinch, line 69: Is it still possible to specify a callback server url in this field with this new type?

Dovchik: @asein-sinch Do you have an example of a value which can be send here because now I'm confused what should go there

Dovchik: Okay, so it can be either json object, or json string, is that right?

Dovchik: Now it takes either string or object

asein-sinch: Yes, when performing a custom call, you can either specify the behaviour inline or specify the callback server url that will process the PIE, which seems to be the most usual way according to the comment.

tests/Sinch.Tests/e2e/Voice/CalloutTests.cs
```diff
@@ -83,7 +83,8 @@ public async Task CustomRequest()
                     .Parse("{\"action\":{\"name\":\"connectPstn\",\"number\":\"46000000001\",\"maxDuration\":90}}")!
                     .AsObject()!,
                 Ace = JsonNode.Parse("{}")!.AsObject(),
-                Pie = "https://your-application-server-host/application"
+                Pie = JsonNode
```
asein-sinch, line 86: Can you also provide an example for the case where the value is a URL string ?

asein-sinch: Thanks
Would the previous line still be valid when a user wants to use a simple string? Or do they have to write `Pie = JsonNode.Parse("https://your-application-server-host/application")`

Dovchik: yeap, JsonNode has an implicit converstion from string.

asein-sinch (approved):

PR #62, opened 2024-04-24
Feat/voice add webhook validation method

Dovchik:
Also, removing support of basic auth in favor of Application Signed, which was the default anyway. For tests merge of https://github.com/sinch/doppelganger/pull/19 is required. 
Also, I'm planning to remove basic auth for verification as well.

### DIFF
--- a/src/Sinch/SinchClient.cs
+++ b/src/Sinch/SinchClient.cs
@@ -109,13 +109,8 @@ namespace Sinch
         /// <param name="appKey"></param>
         /// <param name="appSecret"></param>
         /// <param name="callingRegion">See <see cref="CallingRegion" />. Defaults to <see cref="CallingRegion.Global" /></param>
-        /// <param name="authStrategy">
-        ///     Choose which authentication to use.
-        ///     Defaults to Application Sign request and it's a recommended approach.
-        /// </param>
         /// <returns></returns>
-        public ISinchVoiceClient Voice(string appKey, string appSecret, CallingRegion? callingRegion = null,
-            AuthStrategy authStrategy = AuthStrategy.ApplicationSign);
+        public ISinchVoiceClient Voice(string appKey, string appSecret, CallingRegion? callingRegion = null);
     }
 
     public class SinchClient : ISinchClient
@@ -271,8 +266,7 @@ namespace Sinch
 
         /// <inheritdoc />
         public ISinchVoiceClient Voice(string appKey, string appSecret,
-            CallingRegion? callingRegion = default,
-            AuthStrategy authStrategy = AuthStrategy.ApplicationSign)
+            CallingRegion? callingRegion = default)
         {
             if (string.IsNullOrEmpty(appKey))
                 throw new ArgumentNullException(nameof(appKey), "The value should be present");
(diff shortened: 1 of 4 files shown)


src/Sinch/SinchClient.cs
```diff
@@ -271,27 +266,22 @@ public ISinchAuth Auth
 
         /// <inheritdoc />
         public ISinchVoiceClient Voice(string appKey, string appSecret,
-            CallingRegion? callingRegion = default,
-            AuthStrategy authStrategy = AuthStrategy.ApplicationSign)
+            CallingRegion? callingRegion = default)
```
JPPortier, line 269: Not related to this PR, but:
- `SmsHostingRegion` for SMS (https://github.com/sinch/sinch-sdk-dotnet/blob/main/src/Sinch/SinchOptions.cs#L32)
- `ConversationRegion` for Conversation (https://github.com/sinch/sinch-sdk-dotnet/blob/main/src/Sinch/SinchOptions.cs#L38)
- `CallingRegion` for Voice.

`SmsRegion` & `ConversationRegion` & `VoiceRegion` seems better to provide same naming convention inside this one but also across SDKs

src/Sinch/SinchClient.cs
```diff
@@ -271,27 +266,22 @@ public ISinchAuth Auth
...
 
             if (string.IsNullOrEmpty(appSecret))
                 throw new ArgumentNullException(nameof(appSecret), "The value should be present");
 
-            ISinchAuth auth;
-            if (authStrategy == AuthStrategy.ApplicationSign)
-                auth = new ApplicationSignedAuth(appKey, appSecret);
-            else
-                auth = new BasicAuth(appKey, appSecret);
+            ISinchAuth auth = new ApplicationSignedAuth(appKey, appSecret);
```
asein-sinch, line 277: Maybe you can declare the type as `ApplicationSignedAuth` directly to avoid the cast later on as you are now supporting only this kind of authentication

Dovchik: nah, http use interface

asein-sinch: I don't understand the link with http here. Can you elaborate? To me, it's related to the `SinchVoiceClient` constructor parameters only. Did I miss something specific to cs?

Dovchik: Http class constructed below depends on `ISinchAuth`

asein-sinch: Thanks for the precision, I was focusing on the modified code

tests/Sinch.Tests/Voice/HooksTests.cs
```diff
@@ -0,0 +1,38 @@
...
+using FluentAssertions;
+using Microsoft.Extensions.Primitives;
+using Xunit;
+
+namespace Sinch.Tests.Voice
+{
+    public class HooksTests
+    {
+        [Fact]
+        public void ValidateRequest()
```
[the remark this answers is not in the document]
Dovchik: More tests added!

JPPortier (approved):

PR #68, opened 2024-04-29
Refactor/unify oneofs

Dovchik:
Unify oneof to be interface typed in a project where it makes sense

Refactor verification to specific methods

Split VerificationReport and VerificationStatus response types

Before merge should be merged: https://github.com/sinch/doppelganger/pull/22

### DIFF
--- a/src/Sinch/SMS/Inbounds/BinaryInbound.cs
+++ b/src/Sinch/SMS/Inbounds/BinaryInbound.cs
@@ -5,9 +5,6 @@ namespace Sinch.SMS.Inbounds
 {
     public class BinaryInbound : IInbound
     {
-        [JsonInclude]
-        public SmsType Type { get; private set; } = SmsType.Binary;
-
         /// <summary>
         ///     The ID of this inbound message.
         /// </summary>


src/Sinch/SMS/Inbounds/BinaryInbound.cs
```diff
@@ -0,0 +1,90 @@
+using System;
+using System.Text.Json.Serialization;
+
+namespace Sinch.SMS.Inbounds
+{
+    public class BinaryInbound : IInbound
+    {
+        [JsonInclude]
+        public SmsType Type { get; private set; } = SmsType.Binary;
```
JPPortier, line 9: _(See comment for SmsInbound)_

tests/Sinch.Tests/e2e/Verification/ReportVerificationTests.cs
```diff
@@ -54,26 +50,20 @@ public async Task ReportFlashCallByIdentity()
...
-                            Amount = 0.42,
-                            CurrencyId = "EUR"
-                        },
-                        TerminationPrice = new PriceDetail()
-                        {
-                            Amount = 0.11,
-                            CurrencyId = "EUR"
-                        },
-                        BillableDuration = 40
+                        Type = IdentityType.Number,
```
JPPortier, line 63: Not related to this PR, just a thought about Identity.
Currently only number is supported.
What do you think about creating something like a dedicated `NumberIdentity` class to simplify interface with this type ?
e.g. for StartVerification request:
```
var identity = new NumberIdentity("+123456789");
var request = .... <identity>
```
It will simplify type usage, hide the protocol dedicated value and in case of new identity type will be supported, it won't cause any breaking change

src/Sinch/SMS/Inbounds/BinaryInbound.cs
Dovchik: Yeah, makes sense, I will hide the prop

tests/Sinch.Tests/e2e/Verification/ReportVerificationTests.cs
Dovchik: yeah, makes sense, I'll add that too after release

JPPortier (approved):

PR #69, opened 2024-04-29
fix: get call model adjust field types

Dovchik:
Fix models for get call, see mock before checking out: https://github.com/sinch/doppelganger/pull/23#issue-2269359344


asein-sinch (approved):
LGTM

PR #71, opened 2024-05-13
hotfix(voice, auth): handle zero byte array in app signature

Dovchik:
add and use different url for application management


650elx (approved):
LGTM

PR #73, opened 2024-05-27
feat: add support for fax emails

Dovchik:
[Depends on](https://github.com/sinch/doppelganger/pull/28)


src/Sinch/Fax/Emails/Emails.cs
```diff
@@ -0,0 +1,308 @@
...
+        {
+            _logger = loggerAdapter;
+            _http = httpClient;
+            _projectId = projectId;
+            _baseAddress = baseAddress;
+            _apiBasePath = new Uri(baseAddress, $"/v3/projects/{projectId}/emails");
+        }
+
+        /// <inheritdoc />
+        public Task<ListEmailsResponse<string>> ListForNumber(string serviceId, string phoneNumber, int? pageNumber,
```
asein-sinch, line 138: As this request uses the `/services` path, it would make more sense to implement it the `Services` class and call it from there as it is exposed twice in the specification

src/Sinch/Fax/Emails/Emails.cs
```diff
@@ -0,0 +1,308 @@
...
+            if (pageSize.HasValue)
+            {
+                queryString.Add("pageSize", pageSize.Value.ToString());
+            }
+
+            uriBuilder.Query = queryString.ToString();
+            return _http.Send<ListEmailsResponse<string>>(uriBuilder.Uri, HttpMethod.Get, cancellationToken);
+        }
+
+        // NOTE!: page number currently do nothing ???
```
asein-sinch, line 163: It seems that the parameter name is `page` and not `pageNumber` to make it work

src/Sinch/Fax/Emails/Emails.cs
```diff
@@ -0,0 +1,308 @@
...
+            ExceptionUtils.CheckEmptyString(nameof(email), email);
+
+            var uriBuilder = new UriBuilder(_apiBasePath);
+
+            uriBuilder.Path += $"/{email}/numbers";
+
+            var queryString = HttpUtility.ParseQueryString(string.Empty);
+            if (pageNumber.HasValue)
+            {
+                queryString.Add("pageNumber", pageNumber.Value.ToString());
```
asein-sinch, line 278: Here too the parameter needs to be updated to `page`

tests/Sinch.Tests/e2e/Fax/EmailsTests.cs
```diff
@@ -0,0 +1,156 @@
...
+                {
+                    "+12015555554"
+                }
+            });
+        }
+
+        [Fact]
+        public async Task ListEmailsPlain()
+        {
+            var response = await FaxClient.Emails.ListForNumber("01HXGS1GE2SXS6HKQDMPYM1JHY", "+12015555554", pageNumber: 1,
```
asein-sinch, line 53: Remember to update this parameter here and on the mockserver too

src/Sinch/Fax/Emails/Emails.cs
Dovchik: Yeah, makes sense, I will move it to `Services` in next PR

asein-sinch (approved):

PR #75, opened 2024-06-05
feat: add fax hooks

Dovchik:
Add hooks and example

### DIFF
--- a/examples/WebApi/Controllers/HandleFaxEventController.cs
+++ b/examples/WebApi/Controllers/HandleFaxEventController.cs
@@ -1,0 +1,27 @@
+using Microsoft.AspNetCore.Mvc;
+using Sinch.Fax.Hooks;
+
+namespace WebApiExamples.Controllers
+{
+    [ApiController]
+    [Route("fax")]
+    public class HandleFaxEventController : ControllerBase
+    {
+        [HttpPost]
+        [Route("event")]
+        public IActionResult HandleEvent([FromBody] IFaxEvent faxEvent)
+        {
+            switch (faxEvent)
+            {
+                case IncomingFaxEvent incomingFaxEvent:
+                    break;
+                case CompletedFaxEvent completedFaxEvent:
+                    break;
+                default:
+                    return BadRequest();
+            }
+
+            return BadRequest();
+        }
+    }
+}
--- a/examples/WebApi/Controllers/HandleFaxEventController.cs
+++ b/examples/WebApi/Controllers/HandleFaxEventController.cs
@@ -1,4 +1,6 @@
+using System.Text.Json;
 using Microsoft.AspNetCore.Mvc;
+using Sinch.Fax.Faxes;
 using Sinch.Fax.Hooks;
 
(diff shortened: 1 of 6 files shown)


examples/WebApi/Controllers/HandleFaxEventController.cs
```diff
@@ -0,0 +1,27 @@
...
+    [Route("fax")]
+    public class HandleFaxEventController : ControllerBase
+    {
+        [HttpPost]
+        [Route("event")]
+        public IActionResult HandleEvent([FromBody] IFaxEvent faxEvent)
+        {
+            switch (faxEvent)
+            {
+                case IncomingFaxEvent incomingFaxEvent:
```
asein-sinch, line 16: As a SDK user, I think it would be interesting to see how an event is handled, both as application/json and as multipart/form-data. Can you extend this example?

src/Sinch/Fax/Hooks/IFaxEvent.cs
```diff
@@ -0,0 +1,102 @@
...
+    public interface IFaxEvent
+    {
+        /// <summary>
+        ///     The different events that can trigger a webhook
+        /// </summary>
+        [JsonPropertyName("event")]
+        public FaxEventType Event { get; }
+    }
+
+    public abstract class GenericFaxEvent
```
asein-sinch, line 20: Unless I'm mistaken, it seems that some properties are not declared to support an event with application/json content. I would have expected to see `file` and `fileType`

src/Sinch/Fax/Hooks/IFaxEvent.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            if (type == FaxEventType.CompletedFax.Value)
+            {
+                return elem.Deserialize<CompletedFaxEvent>(options);
+            }
+
+            throw new JsonException($"Failed to match verification method object, got {descriptor.Name}");
+        }
+
+        public override void Write(Utf8JsonWriter writer, IFaxEvent value, JsonSerializerOptions options)
```
asein-sinch, line 86: I'm wondering whether it's useful to keep this method. WDYT?

examples/WebApi/Controllers/HandleFaxEventController.cs
asein-sinch: Yes, I see 2 of them (both valid for the 2 content types)
 - store an incoming fax
 - keep a record of sent faxes

src/Sinch/Fax/Hooks/IFaxEvent.cs
Dovchik: It's required by `JsonConverter` class to be declared, so it should be present, so it's better to implement it anyway - it can be useful for someone in theory.

Dovchik:
Extended examples. Added form data handle. I'm not able to provide a class in SDK directly as it will require to import heavy `AspNetCore` dependency which I don't want to do

asein-sinch (approved):

PR #76, opened 2024-06-11
feat: add support for sending base64 files

Dovchik:


src/Sinch/Fax/Faxes/FaxesClient.cs
```diff
@@ -140,9 +140,20 @@ private class SendFaxResponse
...
 
-            if (request.ContentUrl?.Any() == true)
+            var sendingContentUrls = request.ContentUrl?.Any() == true;
+            var sendingBase64Files = request.Files?.Any() == true;
+            if (sendingContentUrls || sendingBase64Files)
             {
-                _loggerAdapter?.LogInformation("Sending fax with content urls...");
+                if (sendingContentUrls)
+                {
+                    _loggerAdapter?.LogInformation("Sending fax with content urls...");
```
asein-sinch, line 149: If both ContentUrl and Files are set, you will print 2 logs, but in the code above when you send the request with form-data, there is no such log in case a ContentUrl is defined. This makes the behaviour inconsistent between the 2 ways of sending a fax.

Dovchik: `SendFaxRequest` contructors are mutually exclusive, meaning only `Files` or `ContentUr`l or `File` can be set one at a time, thus it's impossible to hit two log entries

asein-sinch: But the API allows to send a contentUrl along with a base64 file or a file attachment. Is it intended to not support this behavior?

Dovchik: Yes, I was thinking the cases are mutually exclusive. Now I need to rethink the design and consult with api team

Dovchik: According to the response from api team, we can send `contentUrl` along with `base64` files or `formData` files. But `formdata` and `base64` ways are mutually exclusive, I've updated the request model to handle that

src/Sinch/Fax/Faxes/FaxesClient.cs
asein-sinch, line 1: Waiting for feedback from the product team to know what is the preferred way to send a fax with contentUrl only

asein-sinch (approved):

PR #77, opened 2024-06-12
fea: use int for ttl

Dovchik:


src/Sinch/Conversation/Messages/Send/SendMessageRequest.cs
```diff
@@ -154,11 +157,34 @@ public override string ToString()
...
     }
+
+    public class TimeToLiveConverter : JsonConverter<int?>
+    {
+        public override int? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            var str = reader.GetString();
+            if (str is null) return null;
+            str = str.Trim();
+            var val = str.TakeWhile(x => x == 's').ToString();
```
asein-sinch, line 175: This condition seems wrong: if the value is `10s`, it will return an empty string as the condition will fail when reading the first character.
Cf doc: https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.takewhile?view=net-8.0 => "Returns elements from a sequence as long as a specified condition is true, and then skips the remaining elements."

asein-sinch, line 175: And looking at the string conversion, I can't see an override of ToString() for IEnumerable so it would use the Object.ToString() et return the FQTN of your IEnumerable instance.
But I see a [ToArray() method](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.toarray?view=net-8.0#system-linq-enumerable-toarray-1(system-collections-generic-ienumerable((-0)))) that would be useful before creating a new string from it.

Dovchik: ah yeah good catch, in test there is no reads of values as well as no API endpoint returns ttl anyware

tests/Sinch.Tests/Conversation/SendMessageTests.cs
```diff
@@ -500,5 +503,23 @@ public async Task SendAllParams()
...
+
+        [Theory]
+        [InlineData("300s", 300)]
+        [InlineData("  300s ", 300)]
+        [InlineData(null, null)]
+        [InlineData("", null)]
+        [InlineData("   ", null)]
+        public void DeserializeTll(string actual, int? expected)
+        {
+            // NOTE: api doesn't return TTL for deserialization anywhere
```
asein-sinch, line 515: In this case, does it make sense to implement a function for it? Can the `Read` function just return null?

asein-sinch: I would be sad for a customer to build the request as a JSON object to then deserialize it and use the SDK just for sending the request. But let's keep it, the implementation looks fine now 🙌

src/Sinch/Conversation/Messages/Send/SendMessageRequest.cs
```diff
@@ -154,11 +157,34 @@ public override string ToString()
...
+            var val = new string(str.TakeWhile(char.IsNumber).ToArray());
+            if (int.TryParse(val, out int result))
+            {
+                return result;
+            }
+
+            return null;
+        }
+
+        public override void Write(Utf8JsonWriter writer, int? value, JsonSerializerOptions options)
```
asein-sinch, line 184: Can you also add a test for this one, in order to cover the case where the TTL is null (the e2e only covers the nominal case)? And maybe check if the value is in the correct range.

Dovchik: ttl null is checked in tests above as well, `WithJson` matcher makes a deep equal comparison, so if `ttl` is present, there will be an error. And any extended validation for other fields then obvious missing id is not supported by sdk

asein-sinch: Let's keep it this way then

asein-sinch (approved):

PR #78, opened 2024-06-17
Feat/add accept language header for sms verification

Dovchik:
Update models to latest

### DIFF
--- a/src/Sinch/Verification/Start/Request/StartCalloutVerificationRequest.cs
+++ b/src/Sinch/Verification/Start/Request/StartCalloutVerificationRequest.cs
@@ -10,5 +10,10 @@ namespace Sinch.Verification.Start.Request
         /// </summary>
         [JsonInclude]
         public override VerificationMethodEx Method { get; } = VerificationMethodEx.Callout;
+
+        /// <summary>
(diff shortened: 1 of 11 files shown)


asein-sinch (approved):

src/Sinch/Verification/Start/Request/StartCalloutVerificationRequest.cs
```diff
@@ -10,5 +10,10 @@ public class StartCalloutVerificationRequest : StartVerificationRequestBase
         /// </summary>
         [JsonInclude]
         public override VerificationMethodEx Method { get; } = VerificationMethodEx.Callout;
+
+        /// <summary>
+        ///     Text-To-Speech engine setting. A language-region identifier according to IANA. Only a subset of those identifiers is accepted.
+        /// </summary>
+        public string? Locale { get; set; }
```
asein-sinch, line 17: This doesn't seem to be related to the PR scope.
Keep in mind that this property is part of an object. A similar object is defined in the callback (with a different description) and states that the upper element "speech" can define various properties.

asein-sinch (approved):
Approved with a point of vigilance about a divergence from the Java SDK

PR #81, opened 2024-08-01
Feat/add methods to root of numbers

Dovchik:
Add available and active numbers to root of Numbers client.
Leaving added methods as obsolete in own tag for backward compatibility.

### DIFF
--- a/src/Sinch/Numbers/Active.cs
+++ b/src/Sinch/Numbers/Active.cs
@@ -25,6 +25,8 @@ namespace Sinch.Numbers
         /// <param name="request"></param>
         /// <param name="cancellationToken"></param>
         /// <returns></returns>
+        [Obsolete(
+            $"This method is obsolete, use {nameof(ISinchNumbers)}.{nameof(Numbers.List)} instead.")]
         Task<ListActiveNumbersResponse> List(ListActiveNumbersRequest request,
             CancellationToken cancellationToken = default);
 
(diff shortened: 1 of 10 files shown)


asein-sinch (commented):
Can you also update the tests to reflect the deprecation of the `Available` and `Active` domains?

src/Sinch/Numbers/Active.cs
```diff
@@ -91,7 +101,9 @@ public ActiveNumbers(string projectId, Uri baseAddress,
             _http = http;
         }
 
-        public Task<ListActiveNumbersResponse> List(ListActiveNumbersRequest request, CancellationToken cancellationToken = default)
+
+        public Task<ListActiveNumbersResponse> List(ListActiveNumbersRequest request,
```
asein-sinch, line 105: I'm curious about why the formatting changed as there is no configuration change in this PR. Is it linked to the IDE? I just want to make sure this is intentional and won't be reverted by an auto-formatting in another PR.

src/Sinch/Numbers/Numbers.cs
```diff
@@ -25,6 +35,42 @@ public interface ISinchNumbers
         ///     release numbers from projects, or list all numbers assigned to a project.
         /// </summary>
         public ISinchNumbersActive Active { get; }
```
asein-sinch, line 37: You may also mark this as obsolete before changing the visibility in the next major

src/Sinch/Numbers/Active.cs
Dovchik: Auto formatted by IDE agnostic `dotnet format`

asein-sinch (approved):

PR #82, opened 2024-08-23
Feat/return test and check unicode test

Dovchik:
Also, returns the test, see https://tickets.sinch.com/browse/DEVEXP-491

### DIFF
--- a/tests/Sinch.Tests/Core/HttpTests.cs
+++ b/tests/Sinch.Tests/Core/HttpTests.cs
@@ -19,6 +19,12 @@ namespace Sinch.Tests.Core
         private readonly ISinchAuth _tokenManagerMock;
         private readonly MockHttpMessageHandler _httpMessageHandlerMock;
 
+        private KeyValuePair<string, string>[] _expiredHeader = new KeyValuePair<string, string>[]
+        {
+            new("www-authenticate",
+                "Bearer error=\"invalid_token\", error_description=\"Jwt expired at 2024-07-08T22:12:28Z\", error_uri=\"https://tools.ietf.org/html/rfc6750#section-3.1\"")
+        };
+
         public HttpTests()
         {
             _tokenManagerMock = Substitute.For<ISinchAuth>();
@@ -39,7 +45,7 @@ namespace Sinch.Tests.Core
             var uri = new Uri("http://sinch.com/items");
             _httpMessageHandlerMock.Expect(HttpMethod.Get, uri.ToString())
                 .WithHeaders("Authorization", "Bearer first_token")
-                .Respond(HttpStatusCode.Unauthorized);
+                .Respond(HttpStatusCode.Unauthorized, _expiredHeader, (HttpContent)null);
             _httpMessageHandlerMock.Expect(HttpMethod.Get, uri.ToString())
                 .WithHeaders("Authorization", "Bearer second_token")
                 .Respond(HttpStatusCode.OK);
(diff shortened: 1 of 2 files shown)


tests/Sinch.Tests/Core/HttpTests.cs
```diff
@@ -191,5 +196,99 @@ public async Task AddOwnHeaders()
...
+        {
+            _tokenManagerMock.GetAuthToken(Arg.Any<bool>())
+                .Returns("first_token");
+
+            var uri = new Uri("http://sinch.com/items");
+
+            // first token expires
+            _httpMessageHandlerMock.Expect(HttpMethod.Get, uri.ToString())
+                .WithHeaders("Authorization", "Bearer first_token")
+                .WithContent(JsonSerializer.Serialize("😼"))
```
650elx, line 284: Nice selection of emojis 👍

650elx (approved):

tests/Sinch.Tests/Core/HttpTests.cs
```diff
@@ -191,5 +196,99 @@ public async Task AddOwnHeaders()
 
             _httpMessageHandlerMock.VerifyNoOutstandingExpectation();
         }
+
+        [Fact]
+        public async Task UnauthorizedIfExpiredHeaderIsNotPresent()
```
asein-sinch, line 201: The test name needs to highlight the fact that no call to the auth server will be made to request a new token as per new logic introduced in the code above.

tests/Sinch.Tests/Core/HttpTests.cs
```diff
@@ -191,5 +196,99 @@ public async Task AddOwnHeaders()
...
+                .Respond(HttpStatusCode.Unauthorized);
+
+            var httpClient = new HttpClient(_httpMessageHandlerMock);
+            var http = new Http(_tokenManagerMock, httpClient, null, new SnakeCaseNamingPolicy());
+            Func<Task<EmptyResponse>> op1 = () => http.Send<EmptyResponse>(uri, HttpMethod.Get);
+
+            await op1.Should().ThrowAsync<SinchApiException>();
+        }
+
+        // This test is testing a positive scenario when server was running idle - without doing requests to sinch api -
```
asein-sinch, line 220: See remarks from https://github.com/sinch/sinch-sdk-dotnet/pull/79#discussion_r1675947673, in particular about the server running idle: the fact a token expires has no link with the server activity.
Maybe a reference with the ticket number will help the other team members to follow what this PR is about.

asein-sinch (changes requested):

tests/Sinch.Tests/Core/HttpTests.cs
Dovchik: I think it's better just to add token

Dovchik:
Updated PR description

asein-sinch (commented):
Great job in making sure to no send more requests to the auth server when the user sends invalid credentials

tests/Sinch.Tests/Core/HttpTests.cs
```diff
@@ -191,5 +196,102 @@ public async Task AddOwnHeaders()
...
+            _httpMessageHandlerMock.Expect(HttpMethod.Get, uri.ToString())
+                .WithHeaders("Authorization", "Bearer first_token")
+                .Respond(HttpStatusCode.Unauthorized);
+
+            var httpClient = new HttpClient(_httpMessageHandlerMock);
+            var http = new Http(_tokenManagerMock, httpClient, null, new SnakeCaseNamingPolicy());
+            Func<Task<EmptyResponse>> op1 = () => http.Send<EmptyResponse>(uri, HttpMethod.Get);
+
+            await op1.Should().ThrowAsync<SinchApiException>();
+            // only one call, returning cached token, is performed
```
asein-sinch, line 218: This comment is misleading as we are checking what's happening on the network here, and a "call" may make one think about an HTTP call while the call to the tokenManagerMock is a function call.
```suggestion
            // only one function call, to read the token stored in cache, is performed
```
Anyway, this seems to be an edge case as it would be awkward to have a JWT in the cache, that is not expired but that the API will treat as invalid.

PR #85, opened 2024-10-04
Feat: add br region to conversation, extract and test url resolutions

Dovchik:
In this PR, besides adding new `br` region to `ConversationRegions.cs`, I'm extracting url resolution logic from `SinchClient` to it's own class: `UrlResolver`. Tests added in `UrlResolverTests` to be sure urls are correctly resolved: url override takes precedence, and regions are templated correctly into url, if applicable. 

Also, adds `VoiceApplicationManagementUrl` to the list of overrides, as api endpoint is different from Voice, for the case when this override is needed.


650elx (approved):
Looks solid 👌

PR #92, opened 2024-10-25
Fix/deserialization of callback events

Dovchik:
https://tickets.sinch.com/browse/DEVEXP-622
Adds `Conversation.Webhooks.Deserialize.ParseEvent(string)` method to deserialize hooks models.
For `Conversation` unit tests, use the same `JsonSerializationOptions` as inside `ConversationClient` to be sure of serialization consistency
Update `Conversation.Webhooks.Update` method to take first paramater as an Id of of a webhook

### DIFF
--- a/src/Sinch/Conversation/Messages/Message/ChoiceMessage.cs
+++ b/src/Sinch/Conversation/Messages/Message/ChoiceMessage.cs
@@ -45,6 +45,21 @@ namespace Sinch.Conversation.Messages.Message
     /// </summary>
     public class UrlMessage
     {
+        [JsonConstructor]
+        public UrlMessage()
+        {
+        }
+
+        [Obsolete(
+            message:
+            "This method is obsolete and will be removed in a future version. Consider initializing properties directly",
+            error: false)]
+        public UrlMessage(string title, Uri url)
+        {
+            Title = title;
+            Url = url.ToString();
+        }
+
         [JsonPropertyName("title")]
 #if NET7_0_OR_GREATER
         public required string? Title { get; set; }


asein-sinch:
The updates in this PR are going to the right direction, but in the current state, it's difficult to provide an approval:
 - the referenced ticket is too vague about what needs to be done
 - the PR is targeting the main branch which will trigger a new release upon merge. We have seen the difficulty to communicate about a version upgrade in the past, it would be better to cover all the deserialization cases in a single release
 - to updated code is handling only a subset of messages types. Can you clarify the scope of this PR and confirm whether the other types of messages need to be updated or not?
The E2E tests targeting the mocks from our official mockserver are passing because only a few types of messages are present. It would be better to add unit tests for this kind of code update.

src/Sinch/Conversation/Messages/Message/ReplyTo.cs
```diff
@@ -1,13 +1,18 @@
...
     /// </summary>
-    /// <param name="MessageId">The Id of the message that this is a response to</param>
-    public sealed record ReplyTo(string MessageId)
+    public sealed class ReplyTo
     {
+        /// <summary>
+        ///      The Id of the message that this is a response to
+        /// </summary>
+        [JsonPropertyName("message_id")]
+        public string? MessageId { get; set; }
```
JPPortier, line 15: `required` by specs but `?` ?

tests/Sinch.Tests/e2e/Conversation/HooksTests.cs
```diff
@@ -0,0 +1,948 @@
...
+using Sinch.Conversation.Contacts;
+using Sinch.Conversation.Events.EventTypes;
+using Sinch.Conversation.Hooks;
+using Sinch.Conversation.Hooks.Models;
+using Sinch.Conversation.Messages.Message;
+using Sinch.Conversation.Webhooks;
+using Sinch.Numbers.Hooks;
+using Xunit;
+
+namespace Sinch.Tests.e2e.Conversation
```
JPPortier, line 19: to ensure a full e2e test coverage: each webhook test should also been validated against `ISinchConversationWebhooks.ValidateAuthenticationHeader` calls

src/Sinch/Conversation/Messages/Message/ChannelSpecificMessage.cs
```diff
@@ -0,0 +1,190 @@
...
+
+namespace Sinch.Conversation.Messages.Message
+{
+    public class ChannelSpecificMessage
+    {
+        /// <summary>
+        /// The message type.
+        /// </summary>
+        [JsonPropertyName("message_type")]
+#if NET7_0_OR_GREATER
```
asein-sinch, line 15: This should not exist anymore: https://community.sinch.com/t5/Developer-Forum/NET-Framework-Updates-for-NET-SDK/td-p/17668
The related PRs are approved. Either they are merged on main, either this PR targets the related branch.

src/Sinch/Conversation/Messages/Message/ReplyTo.cs
JPPortier: Not changed/fixed ?

src/Sinch/Conversation/Messages/Message/ContactMessage.cs
```diff
@@ -48,62 +50,92 @@ public ContactMessage(TextMessage textMessage)
...
+        [JsonInclude]
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        [JsonPropertyName("product_response_message")]
+        public ProductResponseMessage? ProductResponseMessage { get; internal set; }
 
 
+        [JsonInclude]
+        [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+        [JsonPropertyName("channel_specific_message")]
+        public ChannelSpecificMessage? ChannelSpecificMessage { get; internal set; }
```
JPPortier, line 137: It should be of type `ChannelSpecificContactMessage`: not the same type as with AppMessage field

src/Sinch/Conversation/Messages/Message/ChannelSpecificMessage.cs
asein-sinch: Is there a plan somewhere to know what should go where? The PR https://github.com/sinch/sinch-sdk-dotnet/pull/97 is blocked for the opposite reason and your answer was the opposite too: https://github.com/sinch/sinch-sdk-dotnet/pull/97#issuecomment-2566306897
I'm having difficulties reviewing the PR as they are piling up.

src/Sinch/Conversation/Messages/Message/FallbackMessage.cs
```diff
@@ -1,4 +1,5 @@
 using System.Text;
+using System.Text.Json.Serialization;
 
 namespace Sinch.Conversation.Messages.Message
 {
```
asein-sinch, line 5: RawMessage and Reason also need to be deserialized

src/Sinch/Conversation/Messages/Message/ProductResponseMessage.cs
```diff
@@ -0,0 +1,49 @@
...
+    public class ProductResponseMessage
+    {
+        /// <summary>
+        ///     The selected products.
+        /// </summary>
+        [JsonPropertyName("products")]
+#if NET7_0_OR_GREATER
+       public required List<ProductItem> Products { get; set; }
+#else
+        public List<ProductItem> Products { get; set; } = null!;
```
asein-sinch, line 16: You may also check the properties from ProductItem can be deserialized

src/Sinch/Conversation/Messages/Message/ReplyTo.cs
```diff
@@ -1,13 +1,23 @@
...
+using System.Text.Json.Serialization;
 
 namespace Sinch.Conversation.Messages.Message
 {
     /// <summary>
     ///     If the contact message was a response to a previous App message then this field contains information about that.
     /// </summary>
-    /// <param name="MessageId">The Id of the message that this is a response to</param>
-    public sealed record ReplyTo(string MessageId)
+    public sealed class ReplyTo
```
asein-sinch, line 9: Why this discrepancy with TextMessage?

tests/Sinch.Tests/Conversation/MessagesTests.cs
```diff
@@ -406,19 +408,50 @@ public class OmniMessageTestData : IEnumerable<object[]>
...
+                    Type = ChannelSpecificMessageType.NfmReply,
+                    NfmReply = new WhatsAppInteractiveNfmReply()
+                    {
+                        Name = WhatsAppInteractiveNfmReply.NameEnum.AddressMessage,
+                        Body = "nfm reply body value",
+                        ResponseJson = "{\"key\": \"value\"}"
+                    }
+                }
+            }), options => options.Excluding(x => x.ChannelSpecificMessage.Message.NfmReply.ResponseJson));
+            result!.ChannelSpecificMessage!.Message!.NfmReply!.ResponseJson.ToString().Should()
```
asein-sinch, line 453: This could be avoided if the property was treated as a string

src/Sinch/Conversation/Messages/Message/ChoiceMessage.cs
```diff
@@ -40,14 +43,46 @@ public override string ToString()
...
+        public string Url { get; set; } = null!;
+#endif
+    }
 
     /// <summary>
     ///     Message for triggering a call.
     /// </summary>
-    /// <param name="PhoneNumber">Phone number in E.164 with leading +.</param>
-    /// <param name="Title">Title shown close to the phone number. The title is clickable in some cases.</param>
-    public record CallMessage(string PhoneNumber, string Title);
```
asein-sinch, line 52: Is this a breaking change?

tests/Sinch.Tests/e2e/Conversation/WebhooksTests.cs
```diff
@@ -90,14 +113,106 @@ public async Task Delete()
...
         }
 
         [Fact]
         public async Task Update()
         {
-            var response = await SinchClientMockServer.Conversation.Webhooks.Update(_webhookResponse);
-            response.Should().BeEquivalentTo(_webhookResponse);
+            var response = await SinchClientMockServer.Conversation.Webhooks.Update(new Webhook()
+            {
+                Id = "ID",
```
asein-sinch, line 190: It should not be possible to update the ID

asein-sinch: Ok, I see the path parameters and the body parameters are all at the same place. So the ID will be sent over the wire in the body as well as in the path, right?

asein-sinch: Yes, it could be an issue regarding clarity and consistency:

1. Best Practices: It's cleaner to avoid including the Id in the body when it's already part of the path. This reduces potential confusion about whether the body Id is relevant.

1. Unnecessary Payload: Even though the server ignores it, including the Id in the body adds unnecessary information, which can lead to misunderstandings or errors down the line.

Would you consider removing the Id from the body to simplify the request structure?

src/Sinch/Conversation/Webhooks/CreateWebhookRequest.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// </summary>
+        public WebhookTargetType? TargetType { get; set; }
+
+        /// <summary>
+        ///     The app that this webhook belongs to.
+        /// </summary>
+#if NET7_0_OR_GREATER
+        public required string AppId
+#else
+        public string AppId
```
asein-sinch, line 19: Can you clear the warnings created with this commit "Non-nullable property 'AppId' must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring the property as nullable."

src/Sinch/Conversation/Webhooks/UpdateWebhookRequest.cs
```diff
@@ -0,0 +1,136 @@
...
+                SetFields.Add(nameof(ClientCredentials));
+                _clientCredentials = value;
+            }
+        }
+
+        /// <summary>
+        ///     Gets or sets the ID of the webhook.
+        /// </summary>
+        [JsonIgnore]
+        // cannot use required here as System.Text.Json will throw an exception when serializing
```
asein-sinch, line 64: Can you explain what it means? When I read this comment, it makes me think the Id is considered as optional by the class UpdateWebhookRequest whereas it's a required path parameter.
Also, why do you need to serialize this property as it is not part of the request body?

src/Sinch/Conversation/Webhooks/Webhooks.cs
```diff
@@ -142,28 +142,28 @@ public async Task<IEnumerable<Webhook>> List(string appId, CancellationToken can
...
             var queryString = HttpUtility.ParseQueryString(string.Empty);
-            var propMask = webhook.GetPropertiesMask();
+            var propMask = request.GetPropertiesMask();
             if (!string.IsNullOrEmpty(propMask)) queryString.Add("update_mask", propMask);
             builder.Query = queryString.ToString()!;
 
-            _logger?.LogDebug("Updating a webhook with {id}...", webhook.Id);
-            return _http.Send<Webhook, Webhook>(builder.Uri, HttpMethod.Patch, webhook,
+            _logger?.LogDebug("Updating a webhook with {id}...", request.Id);
+            return _http.Send<UpdateWebhookRequest, Webhook>(builder.Uri, HttpMethod.Patch, request,
```
asein-sinch, line 166: With this update, how do you make sure request.Id is not part of the request body?

src/Sinch/Conversation/Messages/Message/ChoiceMessage.cs
```diff
@@ -40,14 +43,40 @@ public override string ToString()
...
+#endif
+    }
 
     /// <summary>
     ///     Message for triggering a call.
     /// </summary>
     /// <param name="PhoneNumber">Phone number in E.164 with leading +.</param>
     /// <param name="Title">Title shown close to the phone number. The title is clickable in some cases.</param>
-    public record CallMessage(string PhoneNumber, string Title);
+    public sealed record CallMessage(string PhoneNumber, string Title)
```
asein-sinch, line 68: A record is sealed by default. What's the benefit of adding the keyword `sealed` ?
From what I understand, you are introducing `get` and `set` methods to make the immutable-by-default `record` into something mutable: are you sure a `record` is the correct structure for that?
Also, from a developer experience point of view, using this record compared to the other model classes will prevent to use the same object initializer syntax
```
new CallMessage()
{
    PhoneNumber = "phone number value",
    Title = "title value"
}
```

asein-sinch (commented):
I see some new Json property names coming after my comments in the commit https://github.com/sinch/sinch-sdk-dotnet/pull/92/commits/e01ff7d5e4def600ccbce54abd06a8155b8e66c8#diff-cbff10904fe00e8dd2965d1f514a7cc6f5aed98061589bdd2d6bc984a1f5f504

However, I don't see the corresponding unit tests. I tried to do an exhaustive review, but I may have missed some classes. Adding some unit tests with the different variations of AppMessage and ContactMessage would ensure we are covered.

asein-sinch (commented):
> added more types covered and double-checked with code coverage but might still miss something as tool not perfect

Is this report visible somewhere?
I've noticed that you have added examples for ContactMessages, but regarding the AppMessages, I can see only "ContactInfoMessage" and "AppMessageCardMessage" (which should be moved to the folder "Messages" BTW)
Given the content of the CardMessage.cs (in particular the property MediaMessage defined as a MediaCarouselMessage instead of a MediaMessage), I think it would be beneficial to cover the AppMessages too. The current test is passing by luck as the json file only contains the property "url", common to both models.
Also for WhatsApp messages, only 2 types of headers are covered for the Flow (text is covered in the hooks and could also be present in the Messages, but video is missing), but nothing for the payment order details and payment order status (I guess these 2 messages will be handled in DEVEXP-578 so not required in this PR)

src/Sinch/Conversation/Webhooks/Webhooks.cs
```diff
@@ -201,6 +224,49 @@ public bool ValidateAuthenticationHeader(Dictionary<string, StringValues> header
...
+            var jsonResult = json.Deserialize<ICallbackEvent>(_http.JsonSerializerOptions);
+            if (jsonResult == null)
+            {
+                throw new InvalidOperationException("Deserialization of callback event failed");
+            }
+
+            return jsonResult;
+        }
+
+        public async Task<ICallbackEvent> ParseEventAsync(Stream jsonStream,
```
JPPortier, line 257: Just for knowledge about .NET
Why are we providing an async function for a feature which do not perform any network request nor having long time execution ?

tests/Sinch.Tests/Conversation/MessagesTests.cs
JPPortier, line 1: For easier reading/maintenance it could have been better to split tests onto files dedicated to each classes (and this could help to validate code coverage too)
About function ` private static object Message()` (I cannot add a comment onto dedicated line because not part of the PR).
The function is creating a payload with both an `app_message` and `contact_message` fields.
But specs say it is a oneOf, so payload used for validation is not valid

tests/Sinch.Tests/e2e/Conversation/WebhooksTests.cs
JPPortier, line 1: By moving to support e2e based onto cucumber and mockerserver; it will help us to ensure e2e are 100% covered according to features description from test suite.

JPPortier:
> Adding additional unit tests... As for the report it shows AppMessage is covered, but by e2e, not unit tests.

Due to complexity of all schemas managed by Conversation: yes a report about coverage could be very helpful (and I think for all of our SDKs)

src/Sinch/Conversation/Messages/Message/CardMessage.cs
```diff
@@ -12,29 +13,34 @@ public sealed class CardMessage : IOmniMessageOverride
...
         ///     This is an optional description field that is displayed below the title on the card.
         /// </summary>
+        [JsonPropertyName("description")]
         public string? Description { get; set; }
 
 
         /// <summary>
         ///     Gets or Sets MediaMessage
         /// </summary>
+        [JsonPropertyName("media_message")]
```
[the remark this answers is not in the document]
JPPortier: Not sure this was fixed.
Could check/confirm there is not a mix here between
- CardMessageMediaMessage specified for CardMessage https://gitlab.com/sinch/sinch-projects/enterprise-and-messaging/documentation/developer-experience/oas-documentation/-/blob/master/services/conversation/conversation.yaml?ref_type=heads#L2244
VS
- MediaMessageField specified for ContactMessage: https://gitlab.com/sinch/sinch-projects/enterprise-and-messaging/documentation/developer-experience/oas-documentation/-/blob/master/services/conversation/conversation.yaml?ref_type=heads#L2720

Be careful there is also an existing MediaCardMessageField schema defined for ContactMessage
If I'm right: the class for CardMessageMediaMessage schema (the one with only `url` field is not defined

JPPortier:
FI: latest builds are having failures.
e.g. : Missing test file ? https://github.com/sinch/sinch-sdk-dotnet/actions/runs/13180441525/job/36789640267#step:12:1966

src/Sinch/Conversation/Messages/Message/ChannelSpecificMessage.cs
asein-sinch: Is this statement still valid? If I've understood correctly, you mentioned about a version 2.0 for the old version deprecation

asein-sinch (commented):
Please check pending comments that are still not answered. On my side:
 - https://github.com/sinch/sinch-sdk-dotnet/pull/92#discussion_r1946408157
 - https://github.com/sinch/sinch-sdk-dotnet/pull/92#discussion_r1931060688
 - https://github.com/sinch/sinch-sdk-dotnet/pull/92#discussion_r1931070605
 - https://github.com/sinch/sinch-sdk-dotnet/pull/92#discussion_r1935101348

tests/Sinch.Tests/Resources/Conversation/Messages/AppMessage/AppMessageListMessage.json
```diff
@@ -10,11 +10,9 @@
...
               "description": "description value",
               "media": {
-                "media_message": {
-                  "url": "an url value",
-                  "thumbnail_url": "another url",
-                  "filename_override": "filename override value"
-                }
+                "url": "an url value",
+                "thumbnail_url": "another url",
+                "filename_override": "filename override value"
```
JPPortier, line 15: Not seeing serialization test when `ListMessage` contains a `product` in place of a `choice`

src/Sinch/Conversation/Webhooks/Webhooks.cs
```diff
@@ -201,6 +224,49 @@ public bool ValidateAuthenticationHeader(Dictionary<string, StringValues> header
...
+            {
+                throw new InvalidOperationException("Deserialization of callback event failed");
+            }
+
+            return jsonResult;
+        }
+
+        public ICallbackEvent ParseEvent(JsonNode json)
+        {
+            var jsonResult = json.Deserialize<ICallbackEvent>(_http.JsonSerializerOptions);
```
asein-sinch, line 248: When a user implements the endpoint that receives the event from Sinch, their primary need would be to parse the incoming string. But the `JsonSerializerOptions` is a property of the `Http` class and from what I read in the SinchClient class the `Http` class needs a `HttpClient` that is either coming from the options, either created by the  SDK itself.
What's the need to create the `Http` class (and mainly the underlying `HttpClient`) where the only goal is to parse a deserialize a json string?

src/Sinch/Conversation/Messages/Message/ReplyTo.cs
asein-sinch: v2 is another story. Let's try to focus on non-breaking changes and dev exp here.

src/Sinch/Conversation/Messages/Message/ChoiceMessage.cs
```diff
@@ -40,14 +43,67 @@ public override string ToString()
     /// <summary>
     ///     A generic URL message.
     /// </summary>
-    /// <param name="Title"></param>
-    /// <param name="Url"></param>
-    public record UrlMessage(string Title, Uri Url);
+    public class UrlMessage
```
asein-sinch, line 46: Shouldn't there be a constructor with both parameters to keep backward compatibility?

src/Sinch/Conversation/Messages/Message/ChoiceMessage.cs
asein-sinch: > Where did you get the information that records are sealed by default?

Indeed, I misinterpreted the immutability property of a record versus the extensibility. Sorry for the confusion.
Regarding the performance and to not allow inheritance arguments, do you plan to generalize this to all the classes?

src/Sinch/Conversation/Webhooks/Webhooks.cs
asein-sinch: As discussed, will be addressed in another ticket

asein-sinch (approved):

PR #97, opened 2024-12-27
Feat/templates/v1

Dovchik:


src/Sinch/Conversation/TemplatesV1/TemplateChannel.cs
```diff
@@ -0,0 +1,22 @@
+using System.Text.Json.Serialization;
+using Sinch.Core;
+
+namespace Sinch.Conversation.TemplatesV1
+{
+    [JsonConverter(typeof(EnumRecordJsonConverter<TemplateChannel>))]
+    public record TemplateChannel(string Value) : EnumRecord(Value)
+    {
+        public static readonly TemplateChannel Unspecfied = new("UNSPECIFIED");
```
asein-sinch, line 9: Typo: `Unspecified`

asein-sinch (commented):
This PR should not target `main` as it requires modifications from other branches

asein-sinch:
> It's okay, the PR won't land in main until other branches are resolved.

Then the approval will wait as there are TODOs in this PR and I will have to re-review again once the main branch is updated. This is not convenient; if you were targeting the right branch , we could move forward easier.

src/Sinch/Conversation/SinchConversationClient.cs
```diff
@@ -52,6 +52,11 @@ public interface ISinchConversation
 
         /// <inheritdoc cref="ISinchConversationTemplatesV1" />
         ISinchConversationTemplatesV1 TemplatesV1 { get; }
+
+        /// <summary>
+        ///     For internal use, JsonSerializerOption to be utilized for serialization and deserialization of all Conversation models
+        /// </summary>
+        internal JsonSerializerOptions JsonSerializerOptions { get; }
```
asein-sinch, line 59: Why do you need to declare this new property? In the ParseEvent method, the instance from Http is used.

PR #105, opened 2025-04-07
feat: change ExpirationTime to DateTime

Dovchik:
https://tickets.sinch.com/browse/DEVEXP-842

### DIFF
--- a/src/Sinch/Core/UnixTimestampSecondsJsonConverter.cs
+++ b/src/Sinch/Core/UnixTimestampSecondsJsonConverter.cs
@@ -11,9 +11,12 @@ namespace Sinch.Core
         {
             if (reader.TokenType == JsonTokenType.Null)
                 return null;
+            if (reader.TokenType != JsonTokenType.String)
+                throw new JsonException("Expected String token type. Got " + reader.TokenType);
             var value = reader.GetString();
-            if (value == null)
-                throw new JsonException("Expected Unix timestamp in seconds as a string representing a number, got null.");
+            if (string.IsNullOrEmpty(value))
+                throw new JsonException(
+                    "Expected Unix timestamp in seconds as a string representing a number, got an empty string.");
             if (!long.TryParse(value, out var timestamp))
             {
                 throw new JsonException("Failed to parse Unix timestamp as a string to Long.");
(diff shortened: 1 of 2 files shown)


src/Sinch/Core/UnixTimestampJsonConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+            JsonSerializerOptions options)
+        {
+            if (reader.TokenType == JsonTokenType.Null)
+                return null;
+            var value = reader.GetString();
+            if (value == null)
+                throw new JsonException("Expected Unix timestamp as a string.");
+            if (!long.TryParse(value, out var timestamp))
+            {
+                throw new JsonException("Expected Unix timestamp as a number.");
```
asein-sinch, line 19: This is confusing regarding line 16: which format is expected? A number or a string ?

src/Sinch/Core/UnixTimestampJsonConverter.cs
asein-sinch, line 1: Can you add unit tests for serialization / deserialization ?

src/Sinch/Core/UnixTimestampJsonConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+    public sealed class UnixTimestampJsonConverter : JsonConverter<DateTime?>
+    {
+        public override DateTime? Read(ref Utf8JsonReader reader, Type typeToConvert,
+            JsonSerializerOptions options)
+        {
+            if (reader.TokenType == JsonTokenType.Null)
+                return null;
+            var value = reader.GetString();
+            if (value == null)
+                throw new JsonException("Expected Unix timestamp as a string.");
```
matsk-sinch, line 16: suggestion: Expected Unix timestamp as a string representing a number

src/Sinch/Core/UnixTimestampJsonConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+            JsonSerializerOptions options)
+        {
+            if (value == null)
+            {
+                writer.WriteNullValue();
+                return;
+            }
+
+            long unixTimestamp = ((DateTimeOffset)value.Value).ToUnixTimeSeconds();
+            writer.WriteStringValue(unixTimestamp.ToString());
```
asein-sinch, line 35: Would there be an issue if we want to serialize unix timestamp with sub-second precision, such as 123456789.123 ?

asein-sinch: From the backend, I don't think so, but this is a generic converter and this is here for serialization. So we need to be robust against the values that will be provided by the SDK users.

asein-sinch (commented):
Sorry, my turn to forget to submit the review 🙈

PR #106, opened 2025-04-15
Feat/numbers/add callback configuration endpoints

Dovchik:
Closes DEVEXP-857

### DIFF
--- a/src/Sinch/Numbers/Callbacks/SinchNumbersCallbacks.cs
+++ b/src/Sinch/Numbers/Callbacks/SinchNumbersCallbacks.cs
@@ -99,6 +99,11 @@ namespace Sinch.Numbers.Callbacks
 
         public Task<CallbackConfiguration> Update(string hmacSecret, CancellationToken cancellationToken = default)
         {
+            if (string.IsNullOrEmpty(hmacSecret))
+            {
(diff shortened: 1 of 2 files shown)


src/Sinch/Numbers/Callbacks/SinchNumbersCallbacks.cs
```diff
@@ -0,0 +1,115 @@
...
+            var uri = new Uri(_baseAddress, $"v1/projects/{_projectId}/callbackConfiguration");
+            return _http.Send<CallbackConfiguration>(uri, HttpMethod.Get, cancellationToken);
+        }
+
+        public Task<CallbackConfiguration> Update(string hmacSecret, CancellationToken cancellationToken = default)
+        {
+            if (string.IsNullOrEmpty(hmacSecret))
+            {
+                throw new ArgumentNullException(nameof(hmacSecret));
+            }
```
asein-sinch, line 105: The API doesn't forbid to send an empty string, but will return a 404 (not really meaningful as an error, but this is what it is). For this very special case, I think it's ok to do some input validation on our side because of the nature of the data to update.

asein-sinch (approved):

PR #109, opened 2025-04-30
hotfix: split template and omnitemplate

Dovchik:
template version is optional

Marks `OmniTemplateReference` as only `IOmniMessageOverrides`: `Required when using 'omni_channel_override' and 'omni_template' fields.`

### DIFF
--- a/tests/Sinch.Tests/Conversation/ConversationTestBase.cs
+++ b/tests/Sinch.Tests/Conversation/ConversationTestBase.cs
@@ -36,5 +36,10 @@ namespace Sinch.Tests.Conversation
         {
             return JsonSerializer.Deserialize<T>(json, Conversation.JsonSerializerOptions);
         }
+
+        protected string SerializeAsConversationClient<T>(T message)
+        {
+            return JsonSerializer.Serialize(message, Conversation.JsonSerializerOptions);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Sinch/Conversation/Messages/Message/TemplateMessage.cs
```diff
@@ -99,7 +150,6 @@ public override string ToString()
             sb.Append("  LanguageCode: ").Append(LanguageCode).Append("\n");
             sb.Append("  Parameters: ").Append(Parameters).Append("\n");
             sb.Append("  TemplateId: ").Append(TemplateId).Append("\n");
-            sb.Append("  Version: ").Append(Version).Append("\n");
             sb.Append("}\n");
             return sb.ToString();
         }
```
asein-sinch, line 155: Does it make sense to keep the ToString() method in this abstract class as you are not calling it from the subclasses?

asein-sinch (commented):
Can you add a test to cover specifically what has been changed in this PR: TemplateReference being used without a `version` property ?

asein-sinch:
> Does it makes much sense though?

Generally when I take the time to write a comment in a review, this is because I think it makes sense. Let me clarify:
 - this PR is about a fixing a bug introduced by an erroneous version of the specification that made a parameter required while it shouldn't be
 - by adding a test, you ensure the app now works when the field is missing (I've read the code, I know it's working don't worry)
 - if someone updates the OAS in the future and accidentally marks the field as required again, this test will fail and alert us
 - the test captures the correct business logic, regardless of what the spec might say. It becomes a living specification

Bugs can come from automated codegen based on the OAS. Tests are our safety net when specs are wrong or or tooling not working properly.

asein-sinch:
Please refer to the branching strategy for the target branch. Is should not be `main`

asein-sinch (approved):

asein-sinch:
You can merge on main, but be sure to backport the commit on the maintenance branch. I don't think it has been created yet.

PR #114, opened 2025-05-09
Fix/cleanup obsolete methods

Dovchik:
Closes [DEVEXP-864](https://tickets.sinch.com/browse/DEVEXP-864)

### DIFF
--- a/src/Sinch/Numbers/VoiceConfigurations/VoiceConfiguration.cs
+++ b/src/Sinch/Numbers/VoiceConfigurations/VoiceConfiguration.cs
@@ -6,12 +6,12 @@ namespace Sinch.Numbers.VoiceConfigurations
     public abstract class VoiceConfiguration
     {
         /// <summary>
-        /// Gets or Sets Type
+        /// Gets or Sets Type.
+        /// Expected to be set by the API response, or by Subclasses
         /// </summary>
         [JsonPropertyName("type")]
         [JsonInclude]
-        // Expected to be set by the API response, or by Subclasses
-        internal virtual VoiceApplicationType Type { get; set; } = VoiceApplicationType.Rtc;
+        internal abstract VoiceApplicationType Type { get; set; }
 
         /// <summary>
         ///     Timestamp when the status was last updated.


src/Sinch/Numbers/VoiceConfiguration.cs
JPPortier, line 1: Is there a technical constraint to have this file  within root Numbers directory but all VoiceConfiguration related classes defined within the dedicated `VoiceConfigurations` one ?

src/Sinch/Numbers/ScheduledVoiceProvisioning.cs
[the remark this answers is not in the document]
asein-sinch: Do we really want to make this parameter required?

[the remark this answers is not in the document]
JPPortier: 1. @Dovchik About having to support getter/setter themselves: they do not have sense to be presented as part of class because of the derivated class (ScheduledVoiceEstProvisioning, ScheduledVoiceFaxProvisioning, ScheduledVoiceRTCProvisioning) contains by themselves the information about the type by the dedicated class name
2. @asein-sinch @Dovchik Backend is returned systematically the type value (this is a "scheduled" response not a request defaulting to RTC) so we could assume at SDK level this value will be present.... but back to point above: do we want to provide it when class name already contains the information

src/Sinch/Numbers/VoiceConfigurations/VoiceConfiguration.cs
```diff
@@ -1,11 +1,8 @@
 using System;
-using System.Collections.Generic;
 using System.Text.Json;
 using System.Text.Json.Serialization;
-using Sinch.Conversation.Messages.Message;
-using Sinch.Numbers.VoiceConfigurations;
 
-namespace Sinch.Numbers
+namespace Sinch.Numbers.VoiceConfigurations
```
JPPortier, line 5: Changing namespace, then breaking change, will have to be highlighted onto migration guide

src/Sinch/Numbers/VoiceConfiguration.cs
JPPortier: Why not having follow same for ScheduledVoiceProvisioning class ?

src/Sinch/Numbers/VoiceConfiguration.cs
```diff
@@ -23,18 +23,6 @@ public class VoiceConfiguration
...
-        [Obsolete(
-            $"Plain {nameof(VoiceConfiguration)} will become abstract in future versions. Use concrete type of {nameof(VoiceRtcConfiguration)} instead.")]
-        public string? AppId { get; set; }
-
-        [JsonPropertyName("scheduledVoiceProvisioning")]
-        [JsonInclude]
-        [Obsolete($"Will be removed in future versions." +
-                  $" See specific {nameof(ScheduledVoiceRtcProvisioning)}, {nameof(ScheduledVoiceEstProvisioning)}, or {nameof(ScheduledVoiceFaxProvisioning)}" +
-                  $" in corresponding classes: {nameof(VoiceRtcConfiguration)}, {nameof(VoiceEstConfiguration)}, or {nameof(VoiceFaxConfiguration)}.")]
-        public ScheduledVoiceProvisioning? ScheduledVoiceProvisioning { get; internal set; }
```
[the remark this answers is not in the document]
JPPortier: But despite OOO language, the end user no longer have access to the "untyped" getter for sheduledVoiceProvisioning from a VoiceConfiguration instance.
he have to explicitly access to the derivated class (VoiceXXXConfiguration/...) before being able to access to the  ScheduledVoiceProvisioning field)

src/Sinch/Numbers/ScheduledVoiceProvisioning.cs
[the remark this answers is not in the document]
asein-sinch: This is a read-only field, meaning it won't be echoed when someone will try to reused the data structure to send another request to update the number. It's ok to remove it.
Regarding my previous comment, I mixed the request / response structure. The property is indeed always sent back from the server, so for a pure DTO point of view, the update made sense.
Now, regarding the need of presenting the `type` in the request or the response, it's highly questionable in an OO language. As mentioned above, the class already bears the context, so it would simplify the usage to hide this property completely from the user and let the SDK fill it at serialization time.

src/Sinch/Numbers/VoiceConfigurations/VoiceConfiguration.cs
asein-sinch: Not ok: the migration guide is for users who are already on v1.x and need to update to v2.0. We should include the breaking changes and how to update the existing code:
 - constructor changes (with an explanation)
 - removed classes, suggested alternatives and/or why they are no longer needed
 - namespace changes with a clear mapping old -> new
 - behavioral changes such as refresh token logic if part of the new version
 
 The release note serves all audiences (not just upgraders). It should summarize the what and the why of the release. It should of course link clearly "For details on how to migrate from v1.x, see the v2.0 migration guide (link)"

src/Sinch/Numbers/VoiceConfiguration.cs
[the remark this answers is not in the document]
JPPortier: Yes this is the idea, but being able to access to `ScheduledVoiceProvisioning` instance field without having to take care of "parent" class.

i.e.:
```
  fooFunction(VoiceConfiguration voiceConfig) {
            switch (voiceConfig.ScheduledVoiceProvisioning) <<-- 'ScheduledVoiceProvisioning' is not defined at VoiceConfiguration so it is not possible
            {
                case ScheduledVoiceEstProvisioning est:
                    ...
                    break;
                case ScheduledVoiceFaxProvisioning fax:
                    break;
                case ScheduledVoiceRtcProvisioning:
                    break;
                default:
                    throw new ArgumentOutOfRangeException(nameof(voiceConfig));
            }
```

src/Sinch/Numbers/VoiceConfigurations/VoiceConfiguration.cs
```diff
@@ -20,6 +20,10 @@ public abstract class VoiceConfiguration
         [JsonInclude]
         [JsonPropertyName("lastUpdatedTime")]
         public DateTime? LastUpdatedTime { get; internal set; }
+
+        [JsonInclude]
+        [JsonPropertyName("scheduledVoiceProvisioning")]
+        public ScheduledVoiceProvisioning? ScheduledVoiceProvisioning { get; internal set; }
```
JPPortier, line 26: We could talk about this during standup: is using `abstract` could not avoid to implement at parent class feature which should be delegated to inherited classes ?

With current implementation, this mean there is a cycling dependency between parent <-> inherited classes. 
This no longer OO

src/Sinch/Numbers/VoiceConfigurations/VoiceConfiguration.cs
```diff
@@ -34,17 +38,36 @@ public sealed class VoiceConfigurationConverter : JsonConverter<VoiceConfigurati
                 var typeStr = type.GetString();
                 if (typeStr == VoiceApplicationType.Fax.Value)
                 {
-                    return elem.Deserialize<VoiceFaxConfiguration>(options);
+                    var faxConfig = elem.Deserialize<VoiceFaxConfiguration>(options);
+                    if (faxConfig != null)
+                    {
+                        (faxConfig as VoiceConfiguration).ScheduledVoiceProvisioning = faxConfig.ScheduledVoiceProvisioning;
```
JPPortier, line 44: No way to handle the deserialization at inherited class level to avoid to handle all inherited "features" at parent level 🤔  ?

JPPortier:
Great, going to the right direction.

1. In addition we should improve unit tests coverage against what was highlighted previously: a scheduledvoiceprovisioning "type" is not tied to the current configuration type; 
2. What is the status regarding 1.3 released version ? (https://github.com/sinch/sinch-sdk-dotnet/blob/v1.3.0/src/Sinch/Numbers/VoiceConfiguration.cs)

And comparing 1.3 and 2.0: for 2.0 there is a ScheduledVoiceProvisioningConverter defined but was not defined (required?) onto 1.3 version. What is the reason ?

src/Sinch/Numbers/VoiceConfigurations/VoiceConfiguration.cs
rpredescu-sinch: Documented `VoiceConfiguration and ScheduledVoiceProvisioning classes moved to new namespace` migration in MIGRATION_GUIDE.md

rpredescu-sinch:
@JPPortier 
> 1. In addition we should improve unit tests coverage against what was highlighted previously: a scheduledvoiceprovisioning "type" is not tied to the current configuration type;

Added unit tests for ScheduledVoiceProvisioning type not being tied to the current configuration type here:
- https://github.com/sinch/sinch-sdk-dotnet/pull/114/changes#diff-268f57141e73c6d41ae45820efa54db8b2225bf5e0d7376febc8e6b15a24f32dR330

src/Sinch/Conversation/TemplatesV1/CreateTemplateRequest.cs
```diff
@@ -21,8 +21,6 @@ public sealed class CreateTemplateRequest
         [JsonPropertyName("create_time")]
         public DateTime? CreateTime { get; set; }
 
-
-        // TODO: make required when net9 PR merges
```
asein-sinch: This should not be deleted as there is some changes to perform: `default_translation` is a required parameter.
But since the TemplateV1 subdomain is now deprecated, you can in fact  remove everything related to it.

MIGRATION_GUIDE.md
JPPortier: Please create a dedicated ticket about MIGRATION_GUIDE structure.

Content is starting to be a list about points; this list become be very long and won't be easy to readable be end users.

Having sections by domains (Numbers/Conversation/... ) could help them

MIGRATION_GUIDE.md
```diff
@@ -228,3 +231,77 @@ var scheduledProvisioning = new ScheduledVoiceEstProvisioning();
...
 ```
+
+## VoiceConfiguration and ScheduledVoiceProvisioning classes moved to new namespace
+
+The following classes have been moved from the `Sinch.Numbers` namespace to `Sinch.Numbers.VoiceConfigurations`:
+
+- `VoiceConfiguration`
+- `ScheduledVoiceProvisioning`
+- `VoiceConfigurationConverter`
+- `ScheduledVoiceProvisioningConverter`
```
JPPortier: Converters do not have to be public and directly used by end users. They do not need to be listed onto migration guide.

MIGRATION_GUIDE.md
```diff
@@ -228,3 +231,77 @@ var scheduledProvisioning = new ScheduledVoiceEstProvisioning();
...
+```csharp
+var voiceConfig = activeNumber.VoiceConfiguration;
+if (voiceConfig is VoiceRtcConfiguration rtcConfig)
+{
+    // handle RTC configuration
+    var appId = rtcConfig.AppId;
+}
+```
+
+**Serialization behavior:** When serializing `VoiceConfiguration` objects, the SDK automatically includes the correct `type` field in the JSON output based on the concrete type. You don't need to set the type manually.
```
JPPortier: This represent too detailed information for migration guide purpose

MIGRATION_GUIDE.md
JPPortier: 👌 https://sinchenterprise.atlassian.net/browse/DEVEXP-1181

MIGRATION_GUIDE.md
rpredescu-sinch: Hi, @JPPortier!
- I have made converters internal to prevent exposing them to end users
- I have removed converters mention from MIGRATION.md document

Thanks!

MIGRATION_GUIDE.md
```diff
@@ -228,3 +231,77 @@ var scheduledProvisioning = new ScheduledVoiceEstProvisioning();
...
+```
+
+## VoiceConfiguration Type property is now internal
+
+The `Type` property on `VoiceConfiguration` and its derived classes (`VoiceRtcConfiguration`, `VoiceEstConfiguration`, `VoiceFaxConfiguration`) is now `internal`. The same applies to `ScheduledVoiceProvisioning` derived classes.
+
+**Impact:**
+- Code that accessed the `Type` property directly will no longer compile
+- The SDK automatically handles the `type` field during serialization/deserialization
+- Use pattern matching or type checking to determine the concrete type
```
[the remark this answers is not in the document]
rpredescu-sinch: I removed unnecessary detailed information from migration document. Thank you!

MIGRATION_GUIDE.md
rpredescu-sinch: I removed unnecessary detailed information from migration document. Thank you!

PR #119, opened 2025-05-28
Feat(DEVEXP-938)/tests/e2e/verification mock server support

Dovchik:
https://tickets.sinch.com/browse/DEVEXP-938
Separate tasks related to PR
- https://tickets.sinch.com/browse/DEVEXP-944
- https://tickets.sinch.com/browse/DEVEXP-946

### DIFF
--- a/tests/Sinch.Tests.Features/Verification/Webhooks.cs
+++ b/tests/Sinch.Tests.Features/Verification/Webhooks.cs
@@ -3,6 +3,7 @@ using System.Text.Json;
 using System.Threading.Tasks;
 using FluentAssertions;
 using Reqnroll;
+using Sinch.Verification;
 using Sinch.Verification.Common;
 using Sinch.Verification.Hooks;
 
@@ -14,11 +15,13 @@ namespace Sinch.Tests.Features.Verification
         private readonly HttpClient _httpClient = new HttpClient();
         private HttpResponseMessage _verificationRequestResponseMessage;
         private HttpResponseMessage _verificationResultResponse;
+        private ISinchVerificationClient _sinchVerificationClient;
+        private string _rawBody;
 
         [Given(@"the Verification Webhooks handler is available")]
         public void GivenTheVerificationWebhooksHandlerIsAvailable()
         {
-            // TODO: init client for header verification
+            _sinchVerificationClient = Utils.SinchVerificationClient;
         }
 
         [When(@"I send a request to trigger a ""Verification Request"" event")]
@@ -29,17 +32,19 @@ namespace Sinch.Tests.Features.Verification
         }
 
         [Then(@"the header of the Verification event ""Verification Request"" contains a valid authorization")]
-        public void ThenTheHeaderOfTheVerificationEventContainsAValidAuthorization()
+        public async Task ThenTheHeaderOfTheVerificationEventContainsAValidAuthorization()
         {
-            // TODO: implement header verification https://tickets.sinch.com/browse/DEVEXP-944
+            _rawBody = await _verificationRequestResponseMessage.Content.ReadAsStringAsync();
+            _sinchVerificationClient.ValidateAuthenticationHeader(HttpMethod.Post, "/webhooks/verification",
+                _verificationRequestResponseMessage.Headers, _verificationRequestResponseMessage.Content.Headers,
+                _rawBody).Should().BeTrue();
         }
 
         [Then(@"the Verification event describes a ""Verification Request"" event type")]
-        public async Task ThenTheVerificationEventDescribesAEventType()
+        public void ThenTheVerificationEventDescribesAEventType()
         {
             // TODO: schema of oas and api response diverge: https://tickets.sinch.com/browse/DEVEXP-946
-            var raw = await _verificationRequestResponseMessage.Content.ReadAsStringAsync();
-            var verificationEvent = JsonSerializer.Deserialize<VerificationRequestEvent>(raw);
+            var verificationEvent = JsonSerializer.Deserialize<VerificationRequestEvent>(_rawBody);
             verificationEvent.As<VerificationRequestEvent>().Should().BeEquivalentTo(new VerificationRequestEvent
             {
                 Id = "1ce0ffee-c0de-5eed-d00d-f00dfeed1337",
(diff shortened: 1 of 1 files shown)


tests/Sinch.Tests.Features/Verification/Start.cs
```diff
@@ -0,0 +1,176 @@
...
+                Identity = Identity.Number("+17818880008")
+            });
+        }
+
+        [Then(@"the response contains the error details of a Data verification")]
+        public async Task ThenTheResponseContainsTheErrorDetailsOfADataVerification()
+        {
+            var ex = await _dataResponseOp.Should().ThrowAsync<SinchApiException>();
+            ex.Which.Message.Should()
+                .BeEquivalentTo("Forbidden:Seamless verification not available for given destination.");
```
asein-sinch, line 172: The part `Forbidden` is added by the SDK if I understand correctly. Have you also thought about propagating other information from the API response such as the `reference` ?

tests/Sinch.Tests.Features/Verification/Status.cs
```diff
@@ -0,0 +1,121 @@
...
+            {
+                Id = "1ce0ffee-c0de-5eed-d11d-f00dfeed1337",
+                Status = VerificationStatus.Successful,
+                CountryId = "FR",
+                VerificationTimestamp = Helpers.ParseUtc("2024-06-06T09:10:27.7264837Z"),
+                Identity = Identity.Number("+33612345678"),
+                CallComplete = true,
+                Price = new Price()
+                {
+                    BillableDuration = 0,
```
asein-sinch, line 71: Where is this property coming from? The comment says "Present only when termination debiting is enabled", but the property should not be added when not returned by the backend

tests/Sinch.Tests.Features/Verification/Status.cs
```diff
@@ -0,0 +1,121 @@
...
+                        Amount = 0,
+                        CurrencyId = "EUR"
+                    },
+                    VerificationPrice = new PriceDetail()
+                    {
+                        Amount = 0.0205d,
+                        CurrencyId = "EUR"
+                    }
+                }
+            });
```
asein-sinch, line 118: This test doesn't raise any warning when a property is present on the server side but not deserialized by the SDK (`callComplete` in this case).
Can you investigate if there is a way to raise a warning when an unexpected property is dropped by the deserialization? And make this part of the test, so that when the mockserver is updated, the test will fail if there is a new property undetected by the SDK.

tests/Sinch.Tests.Features/Verification/Webhooks.cs
```diff
@@ -0,0 +1,85 @@
...
+    public class Webhooks
+    {
+        private readonly HttpClient _httpClient = new HttpClient();
+        private HttpResponseMessage _verificationRequestResponseMessage;
+        private HttpResponseMessage _verificationResultResponse;
+
+        [Given(@"the Verification Webhooks handler is available")]
+        public void GivenTheVerificationWebhooksHandlerIsAvailable()
+        {
+        }
```
asein-sinch, line 21: A TODO is required here too as you'll need to initialize something to be able to verify the authentication

tests/Sinch.Tests.Features/Verification/Webhooks.cs
```diff
@@ -0,0 +1,85 @@
...
+        public async Task ThenTheVerificationEventDescribesAResultEventType()
+        {
+            var raw = await _verificationResultResponse.Content.ReadAsStringAsync();
+            var resultEvent = JsonSerializer.Deserialize<VerificationResultEvent>(raw);
+            // TODO: schema of oas and api response diverge: https://tickets.sinch.com/browse/DEVEXP-946
+            resultEvent.Should().BeEquivalentTo(new VerificationResultEvent()
+            {
+                Id = "1ce0ffee-c0de-5eed-d00d-f00dfeed1337",
+                Event = "VerificationResultEvent",
+                Method = VerificationMethodEx.Sms,
```
asein-sinch, line 79: I find it strange to have to use the extended version of the enum when the extension is not used. Is there a way to use a literal or to use another enum such as `VerificationMethod.Sms`

tests/Sinch.Tests.Features/Verification/Status.cs
asein-sinch: Can you check if it should be completely removed from the model directly?

tests/Sinch.Tests.Features/Verification/Utils.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public static ISinchVerificationClient SinchVerificationClient =>
+            new SinchClient(null, null, null, options =>
+            {
+                options.ApiUrlOverrides = new ApiUrlOverrides()
+                {
+                    VerificationUrl = "http://localhost:3018",
+                    AuthUrl = "http://localhost:3011",
+                };
+            }).Verification("appKey", "YXBwU2VjcmV0");
```
asein-sinch, line 15: Why do you have to use the base64 version of the secret?

tests/Sinch.Tests.Features/Verification/Status.cs
```diff
@@ -0,0 +1,121 @@
...
+        public void GivenTheVerificationServiceIsAvailable()
+        {
+            _sinchVerificationStatus = Utils.SinchVerificationClient.VerificationStatus;
+            _sinchVerificationStatus.Should().NotBeNull();
+        }
+
+        [When(@"I send a request to retrieve a SMS verification status by its verification ID")]
+        public async Task WhenISendARequestToRetrieveAsmsVerificationStatusByItsVerificationId()
+        {
+            _smsVerification = await _sinchVerificationStatus.GetSmsById("1ce0ffee-c0de-5eed-d00d-f00dfeed1337");
```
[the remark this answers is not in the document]
asein-sinch: I remember this point for `start` and `report`, but not for `status` especially when the input is just an ID and from this information there is no possibility to know what we'll get in return. Can you point me to where this has been discussed?

tests/Sinch.Tests.Features/Verification/Webhooks.cs
asein-sinch: No, I was referring to the Webhook parsing, not the start or report request/response. They should not define a value that doesn't map to a schema for which they are the discriminator

tests/Sinch.Tests.Features/Verification/Status.cs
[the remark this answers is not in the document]
asein-sinch: Yes, the changes from the PR 68 need to be reverted in the next version

tests/Sinch.Tests.Features/Verification/Webhooks.cs
asein-sinch: There are not requests here, only comments 😇
You can suggest an code update as an answer to the comment, or an explanation of why you will keep it like this. The code has been like this for a while, this is not the purpose of this PR to solve everything, but to contemplate possible improvement if not too costly.

PR #145, opened 2026-01-13
Merge #141 and #142 pull requests to v2.0 next

rpredescu-sinch:
Merge #141 and #142 pull requests to v2.0 next.

This brings the following two fixes into v2.0 next:
1. Adds the SDK User-Agent header to OAuth requests to enable tracking and debugging of SDK activity
2. https://github.com/sinch/sinch-sdk-dotnet/issues/139


tests/Sinch.Tests/Core/HttpTests.cs
JPPortier: Just for context: 1.4 tests are based onto `SnakeCaseNamingPolicy.Instance`  (https://github.com/sinch/sinch-sdk-dotnet/pull/141/changes#diff-c8620ff980dd3909be2a50c2fcd3a25658d32fe6fa789186c5a948e2993d23a2R155) 

But here there is a `new SnakeCaseNamingPolicy()`.

Is it expected ?

rpredescu-sinch: Hi, Jean-Pierre!
I inherited it from previous tests, but there is no actual reason not to use `SnakeCaseNamingPolicy.Instance`. Actually, it is even better in terms of memory efficiency to reuse the singleton version instead of creating a new object for each test.
I updated all the tests in `HttpTests.cs` file.
Thank you!

JPPortier: Just a thought, ~~not for this PR~~: if it better to use a singleton; we should make constructor private to enforce singleton usage

JPPortier: Does this could spread too much according to current PR scope ?

rpredescu-sinch: Not at all!
This is the only change we need to bring to enforce singleton, no other code is affected but will prevent accidental future instantiations.
I pushed the refactoring.
Thank you!

JPPortier (approved):

PR #146, opened 2026-01-14
Remove duplicate FaxRegion from SinchOptions

rpredescu-sinch:
Currently, FaxRegion is configured at the SinchOptions level:

```
var options = new SinchOptions
{
    FaxRegion = FaxRegion.Us
};
```

However, SinchFaxConfiguration already has a  Region property:

`public FaxRegion? Region { get; init; }`

Suggested fix:
•	Remove `FaxRegion` property from `SinchOptions`
•	Ensure all usages reference `Region` instead

### DIFF
--- a/src/Sinch/SinchOptions.cs
+++ b/src/Sinch/SinchOptions.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Net.Http;
 using Microsoft.Extensions.Logging;
-using Sinch.Fax;
 using Sinch.SMS;
 
 namespace Sinch
@@ -28,10 +27,6 @@ namespace Sinch
         /// </summary>
         public HttpClientHandlerConfiguration? HttpClientHandlerConfiguration { get; set; }
 
-        /// <summary>
-        ///     Set's the regions for the Fax api.
-        /// </summary>
-        public FaxRegion? FaxRegion { get; set; }
 
         /// <inheritdoc cref="ApiUrlOverrides"/>
         public ApiUrlOverrides? ApiUrlOverrides { get; set; }
--- a/tests/Sinch.Tests/Fax/FaxClientTests.cs
+++ b/tests/Sinch.Tests/Fax/FaxClientTests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using FluentAssertions;
 using Sinch.Fax;
 using Xunit;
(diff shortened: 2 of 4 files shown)


src/Sinch/SinchOptions.cs
```diff
@@ -1,7 +1,6 @@
 using System;
 using System.Net.Http;
 using Microsoft.Extensions.Logging;
-using Sinch.Fax;
 using Sinch.SMS;
```
asein-sinch, line 4: Do you also have a ticket to take care about the SMS options too?

asein-sinch (commented):
Can you also add a test to make sure the value given in the SinchConfiguration is reflecting on the API URL ?

src/Sinch/SinchOptions.cs
JPPortier: Yes and attached to V2.0: DEVEXP-990

JPPortier:
> Can you also add a test to make sure the value given in the SinchConfiguration is reflecting on the API URL ?

Could be part of another ticket to cover all these missing tests (not only for Fax but for all domains)

src/Sinch/SinchOptions.cs
rpredescu-sinch: Hi, @asein-sinch, @JPPortier!
I have added unit tests for `FaxConfiguration` to verify that the `Region` property reflects on the **API URL**. 
The tests cover all available regions and the URL override functionality.
Let me know what do you think!

tests/Sinch.Tests/Fax/FaxClientTests.cs
```diff
@@ -0,0 +1,148 @@
...
+namespace Sinch.Tests.Fax
+{
+    public class FaxClientTests
+    {
+        [Fact]
+        public void FaxConfiguration_ShouldResolveDefaultUrl_WhenNoRegionSpecified()
+        {
+            var config = new SinchClientConfiguration()
+            {
+                SinchUnifiedCredentials = new SinchUnifiedCredentials()
```
asein-sinch: Thanks for adding the tests. I see a lot of repetition here, what is your opinion about using parameterized tests? (expect for the last one where the comment about the region value being discarded is interesting)

rpredescu-sinch: Nice call!
I have simplified unit tests to use parameterized data so now we only have two.
Thank you!

asein-sinch (approved):
Thank you for the updates!

PR #149, opened 2026-01-19
Store SDK version within repository

rpredescu-sinch:
## Summary

The .NET SDK version in `AssemblyInfo.cs` has two issues:

Incorrect version format: displays 4 digits (e.g., 0.1.0.0) instead of 3 digits (e.g., 2.0.0) which is the standard format used by Sinch SDKs

Wrong version in repository: the version is not updated to match the current release (2.0.0), causing users who clone the repository to report incorrect SDK version in logs and User-Agent headers.

While the official NuGet package has the correct version, local builds from the repository source will log activity with an incorrect version.

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,5 +1,5 @@
 mode: ContinuousDelivery
-next-version: 0.1.2
+next-version: 2.0.0
 branches:
   main:
     tag: 'alpha'
--- a/src/Sinch/Sinch.csproj
+++ b/src/Sinch/Sinch.csproj
@@ -6,7 +6,7 @@
         <RepositoryUrl>https://github.com/sinch/sinch-sdk-dotnet</RepositoryUrl>
         <Version>2.0.0</Version>
         <Title>Sinch</Title>
-        <Authors>Volodymyr Lisovskyi</Authors>
+        <Authors>Volodymyr Lisovskyi, Razvan Predescu</Authors>
         <Description>Sinch REST API SDK to work with sms, messaging, numbers, verification and more.</Description>
         <PackageProjectUrl>https://developers.sinch.com/</PackageProjectUrl>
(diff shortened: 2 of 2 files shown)


src/Sinch/Sinch.csproj
```diff
@@ -4,7 +4,7 @@
         <Nullable>enable</Nullable>
         <TargetFrameworks>net8.0;net9.0</TargetFrameworks>
         <RepositoryUrl>https://github.com/sinch/sinch-sdk-dotnet</RepositoryUrl>
-        <Version>0.1.0</Version>
+        <Version>2.0.0</Version>
         <Title>Sinch</Title>
         <Authors>Volodymyr Lisovskyi</Authors>
```
asein-sinch: Don't hesitate to update this property too, you're part of it too now!

asein-sinch (approved):

GitVersion.yml
```diff
@@ -1,5 +1,5 @@
 mode: ContinuousDelivery
-next-version: 0.1.2
```
JPPortier, line 2: It is not clear for me what will do the github action when we will release.
According to: https://gitversion.net/docs/reference/configuration
```
next-version

Allows you to bump the next version explicitly. Useful for bumping main or a feature branch with breaking changes (i.e., a major increment), indicating what the next git tag is going to be.

next-version is not a permanent replacement for git tag and should only be used intermittently. Since version 5.5 GitVersion supports next-version with mode: Mainline and should not be treated as a "base version".

If you are using next-version and are experiencing weird versioning behaviour, please remove it, create a git tag with an appropriate version number on an appropriate historical commit and see if that resolves any versioning issues you may have. 
```

From this config, were we are in `ContinuousDelivery`

So, not a blocker, just we need to take care of the behavior of this plugin/github action during release process

JPPortier (approved):

PR #151, opened 2026-01-20
Schedule provisioning with strongly-typed failure error codes

rpredescu-sinch:
## Summary

This PR changes the `ErrorCodes` property on `ScheduledProvisioning` from `List<string>?` to `IList<FailureCode>?` to provide strongly-typed error codes instead of raw strings.

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -346,7 +346,7 @@ Version 1.*:
 using Sinch.Numbers;
 
 var scheduledProvisioning = activeNumber.SmsConfiguration?.ScheduledProvisioning;
-if (scheduledProvisioning?.ErrorCodes?.Contains("SMS_PROVISIONING_FAILED"))
+if (scheduledProvisioning?.ErrorCodes?.Contains("CAMPAIGN_NOT_AVAILABLE"))
 {
     // handle error
 }
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -14,6 +14,7 @@
 - [Removed obsolete UrlMessage and CallMessage constructors](#removed-obsolete-urlmessage-and-callmessage-constructors)
 - [FaxRegion moved from SinchOptions to SinchFaxConfiguration](#faxregion-moved-from-sinchoptions-to-sinchfaxconfiguration)
 - [Numbers API: Callbacks renamed to CallbackConfiguration](#callbacks-renamed-to-callbackconfiguration)
+- [Numbers API: ScheduledProvisioning.ErrorCodes type changed to IList\<FailureCode\>](#scheduledprovisioningerrorcodes-type-changed-to-ilistfailurecode)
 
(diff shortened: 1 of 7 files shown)


src/Sinch/Numbers/ScheduledProvisioning.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using Sinch.Numbers.Hooks;
```
asein-sinch: `FailureCode` should change namespace. This is not normal that this class depends on `Hooks`

MIGRATION_GUIDE.md
```diff
@@ -335,3 +336,31 @@ Version 2.*:
...
+}
+```
+
+Version 2.*:
+```csharp
+using Sinch.Numbers;
+using Sinch.Numbers.Hooks;
+
+var scheduledProvisioning = activeNumber.SmsConfiguration?.ScheduledProvisioning;
+if (scheduledProvisioning?.ErrorCodes?.Contains(FailureCode.CampaignNotAvailable))
```
asein-sinch, line 361: Nit: can you align the error for both version? Such as using `"CAMPAIGN_NOT_AVAILABLE"` for the v1 ?

tests/Sinch.Tests/Numbers/ActiveNumberTests.cs
```diff
@@ -261,6 +262,49 @@ public void DeserializeActiveNumber()
...
+                ErrorCodes = new List<FailureCode>
+                {
+                    FailureCode.CampaignNotAvailable,
+                    new("CUSTOM_ERROR_CODE")
+                }
+            });
+        }
+
+        [Fact]
+        public void SerializeScheduledProvisioningWithErrorCodes()
```
JPPortier: It do not hurt, but a scheduled provisioning will only be present onto response not requests

src/Sinch/Numbers/ScheduledProvisioning.cs
rpredescu-sinch: Hi, @asein-sinch!
Moved `FailureCode` class to `Sinch.Numbers` namespace because it is used across the Numbers API.
Thank you!

MIGRATION_GUIDE.md
rpredescu-sinch: Now using `CAMPAIGN_NOT_AVAILABLE` for Version 1 error code to align with Version 2 example. Thank you!

tests/Sinch.Tests/Numbers/ActiveNumberTests.cs
rpredescu-sinch: Hi, @JPPortier!
Thank you for catching this! 
I'm removing the test since serializing `ScheduledProvisioning` is not a real use case. 
Best!

asein-sinch (approved):

JPPortier (approved):

PR #153, opened 2026-01-28
Support of E2E tests for SMS webhooks

rpredescu-sinch:
## Summary
.NET SDK runs without failure supporting SMS Webhooks E2E tests.


examples/templates/webhooks/Sms/ServerBusinessLogic.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        _logger.LogInformation("Received recipient report: {Recipient}", reportSms.Recipient);
+        return Task.CompletedTask;
+    }
+
+    private Task HandleRecipientDeliveryReportMms(RecipientDeliveryReportMms report)
+    {
+        _logger.LogInformation("Received recipient report: {Recipient}", report.Recipient);
+        return Task.CompletedTask;
+    }
```
asein-sinch: Nit: for each of the logs, you can also add SMS or MMS for clarity

examples/templates/webhooks/Sms/ServerBusinessLogic.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        _logger.LogInformation("Received binary SMS from {From}", message.From);
+        return Task.CompletedTask;
+    }
+
+    private Task HandleMediaMessage(MediaMessage message)
+    {
+        _logger.LogInformation("Received MMS from {From} with {Count} media items", message.From, message.MessageBody.Media?.Count ?? 0);
+        return Task.CompletedTask;
+    }
```
asein-sinch: For clarity, the methods can be named Handle**Incoming**...() to echo the developer doc.

examples/templates/webhooks/Program.cs
```diff
@@ -0,0 +1,37 @@
...
+
+builder.Services.AddSinchClient(() => new SinchClientConfiguration
+{
+    SinchUnifiedCredentials = new SinchUnifiedCredentials
+    {
+        ProjectId = builder.Configuration["Sinch:ProjectId"]!,
+        KeyId = builder.Configuration["Sinch:KeyId"]!,
+        KeySecret = builder.Configuration["Sinch:KeySecret"]!
+    }
+});
```
asein-sinch, line 30: You can add a note (here and/or in the readme) saying this is not strictly needed in the scope of webhooks validation (only the webhook secret is needed), but it can be useful in the customer's use case is to **answer** to the incoming messages with the SMS API.

examples/templates/webhooks/Sms/ServerBusinessLogic.cs
```diff
@@ -0,0 +1,86 @@
...
+                await HandleBatchDeliveryReportMms(deliveryReportMms);
+                break;
+            case RecipientDeliveryReportSms recipientDeliveryReport:
+                await HandleRecipientDeliveryReport(recipientDeliveryReport);
+                break;
+            case RecipientDeliveryReportMms recipientDeliveryReportMms:
+                await HandleRecipientDeliveryReportMms(recipientDeliveryReportMms);
+                break;
+            default:
+                _logger.LogInformation("Unknown SMS event type: {Type}", smsEvent.GetType());
```
matsk-sinch: Should it be a warning instead of Info?

tests/Sinch.Tests.Features/Sms/Webhooks.cs
```diff
@@ -0,0 +1,189 @@
...
+            {
+                PropertyNamingPolicy = JsonNamingPolicy.SnakeCaseLower
+            });
+            recipientDeliveryReport.Should().BeEquivalentTo(new RecipientDeliveryReportSms
+            {
+                At = Helpers.ParseUtc("2024-06-06T08:17:19.210Z"),
+                BatchId = "01W4FFL35P4NC4K35SMSBATCH9",
+                ClientReference = "client-ref",
+                Code = 0,
+                OperatorStatusName = Helpers.ParseUtc("2024-06-06T08:17:00Z"),
```
matsk-sinch: Since it's a `datetime`, shouldn't be OperatorStatusAt?

examples/templates/webhooks/Sms/ServerBusinessLogic.cs
JPPortier: +10

examples/templates/webhooks/Sms/ServerBusinessLogic.cs
```diff
@@ -0,0 +1,86 @@
...
+        return Task.CompletedTask;
+    }
+
+    private Task HandleMediaMessage(MediaMessage message)
+    {
+        _logger.LogInformation("Received MMS from {From} with {Count} media items", message.From, message.MessageBody.Media?.Count ?? 0);
+        return Task.CompletedTask;
+    }
+
+    private Task HandleBatchDeliveryReport(BatchDeliveryReportSms reportSms)
```
JPPortier: The second function is having just variable named `report`

Better to have a common pattern: always "report" or always "report+\<suffix\>"

(same comment for recipient delivery report)

examples/templates/webhooks/Sms/SinchWebhookFilter.cs
```diff
@@ -0,0 +1,52 @@
...
+/// <summary>
+/// Action filter that validates Sinch SMS webhook HMAC signature before executing the action.
+/// </summary>
+public class SinchWebhookFilter : IAsyncActionFilter
+{
+    private readonly ISmsWebhooks _webhooks;
+    private readonly IConfiguration _configuration;
+    private readonly bool _requireAuthentication;
+
+    public SinchWebhookFilter(ISmsWebhooks webhooks, IConfiguration configuration, bool requireAuthentication = true)
```
JPPortier: > bool requireAuthentication = true

Does this mean by default validation will be performed ?

It should not then easier onboarding onto Sinch integration.

examples/templates/webhooks/Sms/SinchWebhookFilter.cs
```diff
@@ -0,0 +1,52 @@
...
+
+    public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next)
+    {
+        var request = context.HttpContext.Request;
+
+        using var reader = new StreamReader(request.Body, System.Text.Encoding.UTF8);
+        var body = await reader.ReadToEndAsync();
+
+        // Always read and cache the body so controllers can reuse it regardless of auth requirement.
+        context.HttpContext.Items[SinchWebhookConstants.BodyItemKey] = body;
```
JPPortier, line 21: Just a question: the context is a context a request level ? No risk about multi-threading issue and overriding a value here ?

examples/templates/webhooks/Sms/SinchWebhookAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Webhook.Template.Sms;
+
+/// <summary>
+/// Attribute you can apply to controllers or actions to enable Sinch SMS webhook validation.
+/// </summary>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+public class SinchWebhookAttribute : TypeFilterAttribute
+{
+    public SinchWebhookAttribute(bool requireAuthentication = true)
```
JPPortier: > (bool requireAuthentication = true)

Same question about require authentication activated by default: it should be "false" by default

If this parameter is same as the one set onto "SinchWebhookFilter" class; does it need to define default value multiple times ?

examples/templates/webhooks/Sms/WebhooksController.cs
```diff
@@ -0,0 +1,41 @@
...
+/// Ensure valid authentication to handle request.
+/// See https://developers.sinch.com/docs/sms/api-reference/sms/tag/Webhooks/#tag/Webhooks/section/Callbacks
+/// Contact your account manager to configure your callback sending headers validation and
+/// set requireAuthentication to true to validate request from Sinch servers.
+/// See https://developers.sinch.com/docs/numbers/api-reference/numbers/tag/Numbers-Callbacks for
+/// more information.
+/// </summary>
+
+[ApiController]
+[SinchWebhook(requireAuthentication: false)]
```
JPPortier, line 16: Curious if the information about activating authentication or not could be define at a single place only: currently there is 3 different "requireAuthentication" settings.

Can define it at a single place with dedicated doc/pointers about whet does this mean and how to activate it.

examples/templates/webhooks/appsettings.json
```diff
@@ -0,0 +1,17 @@
+﻿{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
+    }
+  },
+  "AllowedHosts": "*",
```
JPPortier: Is it required ?

examples/templates/webhooks/README.md
```diff
@@ -0,0 +1,64 @@
...
+      "WebhookSecret": ""
+    }
+  }
+}
+```
+
+- `Sinch:ProjectId` - your Sinch project id
+- `Sinch:KeyId` - your Sinch API key id
+- `Sinch:KeySecret` - your Sinch API key secret
+- `Sinch:Sms:WebhookSecret` - optional shared secret used to validate incoming webhook signatures (the controller shows how to use this)
```
JPPortier: You could add a '\[...](...)' markdown link to the file

examples/templates/webhooks/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Server URLs and ports
+- Development launch settings (Properties/launchSettings.json) set the local URLs to:
```
JPPortier: Same about markdown link to the file

examples/templates/webhooks/README.md
```diff
@@ -0,0 +1,64 @@
...
+When running under `dotnet run` in development, the app will typically listen on those URLs.
+
+## How to run
+
+Restore and run the project from the `examples/templates/webhooks` folder:
+
+```powershell
+# from repository root or examples/templates/webhooks
+dotnet restore
+dotnet run --project .\Webhook.Template.csproj
```
JPPortier: This command contains an absolute path to current location.

Is ti compliant with comment above about "from root":

> # from repository root or examples/templates/webhooks

examples/templates/webhooks/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+**Authentication / Validating callbacks**
+The controller includes an optional validation step that checks headers and payload against a shared secret (configured at `Sinch:Sms:WebhookSecret`). The example defaults to `ensureValidAuthentication = false` — enable and configure it if you require strict validation of incoming callbacks.
```
JPPortier: Provide too a markdown link to the file to be edited

src/Sinch/SMS/Hooks/TextMessage.cs
```diff
@@ -5,63 +5,49 @@
...
 {
     /// <summary>
-    ///         An inbound message is a message sent to one of your short codes or long numbers from a mobile phone.
-    ///         To receive inbound message callbacks, a URL needs to be added to your REST API.
-    ///         This URL can be specified in your <see href="https://dashboard.sinch.com/sms/api">Dashboard</see>.
+    ///     Text MO (Mobile Originated) - an inbound text SMS message.
     /// </summary>
-    public class IncomingTextSms : IIncomingSms
+    /// <seealso href="https://developers.sinch.com/docs/sms/api-reference/sms/tag/Webhooks/">SMS Webhooks Documentation</seealso>
+    public class TextMessage : IInboundMessage
```
asein-sinch: How much different is this class from the `SmsInbound`? 
And regarding the naming, I think it's interesting to keep the "Inbound" notion in the name to differentiate MO from MT. For instance, in the Conversation API, there is a class `TextMessage` too but it's a message to be sent to the mobile.

src/Sinch/SMS/Hooks/SmsWebhooks.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Collections.Generic;
+using System.Text.Json;
+using Sinch.Logger;
+
+namespace Sinch.SMS.Hooks
+{
+    /// <inheritdoc />
+    internal sealed class SmsWebhooks : ISmsWebhooks
+    {
+        private readonly HmacAuthenticationValidation _authenticationChecker;
```
asein-sinch: Can you check this warning? 
```
The field 'SmsWebhooks._authenticationChecker' is never used
```

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportSms.cs
```diff
@@ -22,7 +25,7 @@ public sealed class DeliveryReport
         ///     The delivery report type.
         /// </summary>
         [JsonPropertyName("type")]
-        public DeliveryReportType? Type { get; set; }
+        public DeliveryReportType? Type { get; set; } = DeliveryReportType.Sms;
```
asein-sinch: Required attribute (this is correct in `BatchDeliveryReportMms`)

src/Sinch/SMS/Hooks/BinaryMessage.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        [JsonPropertyName("received_at")]
+        public required DateTime ReceivedAt { get; set; }
+
+        /// <summary>
+        ///     When the message left the originating device. Only available if provided by operator.
+        ///     Formatted as <see href="https://en.wikipedia.org/wiki/ISO_8601">ISO-8601</see>: YYYY-MM-DDThh:mm:ss.SSSZ.
+        /// </summary>
+        [JsonPropertyName("sent_at")]
+        public DateTime SentAt { get; set; }
```
asein-sinch: Optional value

src/Sinch/SMS/Hooks/BinaryMessage.cs
```diff
@@ -0,0 +1,102 @@
...
+        ///     The Sinch phone number or short code to which the message was sent.
+        /// </summary>
+        [JsonPropertyName("to")]
+        public required string To { get; set; }
+
+        /// <summary>
+        ///     Gets or sets the Type. SMS in binary format.
+        /// </summary>
+        [JsonPropertyName("type")]
+        public required Sinch.SMS.Inbounds.SmsType Type { get; set; }
```
asein-sinch: This could be set to a default value as it is the case for the delivery reports classes, couldn't it?

src/Sinch/SMS/Hooks/BinaryMessage.cs
```diff
@@ -0,0 +1,102 @@
...
+        ///     The UDH header of a binary message HEX encoded. Max 140 bytes together with body.
+        /// </summary>
+        [JsonPropertyName("udh")]
+        public required string Udh { get; set; }
+
+        /// <summary>
+        ///     Returns the string presentation of the object.
+        /// </summary>
+        /// <returns>String presentation of the object.</returns>
+        public override string ToString()
```
asein-sinch: Only this class and `TextMessage` (among the ones implementing `ISmsEvent`) have a `ToString` method. Why these ones and not the others?

src/Sinch/SMS/Hooks/MediaMessage.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <summary>
+    /// Incoming MMS message webhook event
+    /// </summary>
+    public class MediaMessage : IInboundMessage
+    {
+        [JsonPropertyName("type")]
+        public string Type { get; set; } = "mo_media";
+
+        [JsonPropertyName("id")]
+        public string Id { get; set; } = string.Empty;
```
asein-sinch: Why not using the `required` keyword as in `BinaryMessage` ?

src/Sinch/SMS/Hooks/MediaMessageBodyDetails.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// MIME type of the media
+        /// </summary>
+        [JsonPropertyName("content_type")]
+        public string ContentType { get; set; } = string.Empty;
+
+        /// <summary>
+        /// Upload status
+        /// </summary>
+        [JsonPropertyName("status")]
+        public MediaStatus Status { get; set; }
```
asein-sinch: Here you have a warning: either add `required` (preferred) either use a default value.

src/Sinch/SMS/DeliveryReports/RecipientDeliveryReportMms.cs
```diff
@@ -0,0 +1,47 @@
...
+        public string BatchId { get; set; } = string.Empty;
+
+        [JsonPropertyName("client_reference")]
+        public string? ClientReference { get; set; }
+
+        [JsonPropertyName("at")]
+        public DateTime At { get; set; }
+
+        [JsonPropertyName("code")]
+        public int Code { get; set; }
```
asein-sinch: For an `int`, should it have a default value, the same way you set a default `string.Empty` for required string attributes? Or should you use the `required` keyword everywhere?

tests/Sinch.Tests/Sms/WebhooksTests.cs
```diff
@@ -0,0 +1,258 @@
...
+                {
+                    Id = "01XXXXX21XXXXX119Z8P1XXXXX",
+                    Body = "This is a test message.",
+                    From = "16051234567",
+                    To = "13185551234",
+                    OperatorId = "string",
+                    ClientReference = "text-client-ref",
+                    ReceivedAt = System.DateTime.Parse("2022-08-24T14:15:22Z").ToUniversalTime(),
+                    SentAt = System.DateTime.Parse("2022-08-24T14:15:22Z").ToUniversalTime(),
+                    Type = Sinch.SMS.Inbounds.SmsType.Text
```
asein-sinch: The behaviour with `MediaMessage` is different as the type is set in the class itself. Consider aligning the behaviour between these classes (also for `BinaryMessage`)

tests/Sinch.Tests/Sms/WebhooksTests.cs
```diff
@@ -0,0 +1,258 @@
...
+                    Recipient = "+44231235674",
+                    Status = "Dispatched",
+                    At = System.DateTime.Parse("2022-08-30T08:16:08.930Z").ToUniversalTime(),
+                    Operator = "operator",
+                    AppliedOriginator = "applied originator",
+                    ClientReference = "client reference",
+                    Encoding = "encoding",
+                    NumberOfMessageParts = 123,
+                    OperatorStatusAt = System.DateTime.Parse("2022-08-30T08:16:08.150Z").ToUniversalTime(),
+                    Type = "recipient_delivery_report_mms"
```
asein-sinch: Same remark about the necessity to set the type where the class already says what it is.
And on top of that, should you use a string or an existing enum value (as you are doing in the test above with `DeliveryReportType.Mms`) ?

tests/Sinch.Tests/Resources/Sms/Inbounds/InboundText.json
```diff
@@ -0,0 +1,10 @@
+﻿{
+  "type": "mo_text",
+  "id": "in-bound",
+  "from": "+4800000",
+  "to": "+48000001",
+  "body": "hello-world",
+  "received_at": "2019-08-24T14:15:22Z",
+  "udh": "hey"
```
asein-sinch: Is it expected to have extra fields in this schema?

tests/Sinch.Tests/Resources/Sms/Inbounds/InboundBinary.json
asein-sinch, line 1: Maybe I missed it, but I can't see a test using those JSON files (and if it is the case, there is no test file for the Media variant)

tests/Sinch.Tests.Features/Sms/Webhooks.cs
asein-sinch: Good catch 👏
This is indeed a bug from the previous implementation
```
[JsonPropertyName("operator_status_at")]
public DateTime OperatorStatusName { get; set; }  
                              ^^^^
                              Wrong variable name 🙃
```

examples/templates/webhooks/Sms/ServerBusinessLogic.cs
rpredescu-sinch: Hi @matsk-sinch!
Now logging warning instead of information.
Thank you!

tests/Sinch.Tests.Features/Sms/Webhooks.cs
rpredescu-sinch: Thanks for catching this @matsk-sinch and @asein-sinch!
I renamed the property to `OperatorStatusAt`.
Best!

examples/templates/webhooks/Sms/ServerBusinessLogic.cs
rpredescu-sinch: Now using `HandleIncoming...` pattern in method names. Thank you!

examples/templates/webhooks/Sms/ServerBusinessLogic.cs
rpredescu-sinch: Added missing SMS & MMS information to corresponding logs. Thank you!

examples/templates/webhooks/Sms/ServerBusinessLogic.cs
rpredescu-sinch: Hi, @JPPortier!
Renamed variables to generic `report` for simplicity, but the logs are now specifying SMS or MMS in the details.
Thank you!

examples/templates/webhooks/Sms/SinchWebhookFilter.cs
rpredescu-sinch: Hi, @JPPortier!
Yes, `HttpContext.Items` is request scoped. Each HTTP request gets its own `HttpContext` instance, so there's no risk of multi threading issues or values being overridden across requests.
Best!

examples/templates/webhooks/Sms/SinchWebhookAttribute.cs
rpredescu-sinch: Hi, @JPPortier!
I removed the default to prevent hidden assumptions about the behavior. Users must always be explicit: `[SinchWebhook(requireAuthentication: false)]` or `[SinchWebhook(requireAuthentication: true)]`. In our example is false so it works without extra authentication configuration. 
What do you think?

examples/templates/webhooks/Sms/SinchWebhookFilter.cs
rpredescu-sinch: Good catch!
I addressed this below.

> Hi, @JPPortier! I removed the default to prevent hidden assumptions about the behavior. Users must always be explicit: `[SinchWebhook(requireAuthentication: false)]` or `[SinchWebhook(requireAuthentication: true)]`. In our example is false so it works without extra authentication configuration. What do you think?

Thank you!

examples/templates/webhooks/Sms/WebhooksController.cs
rpredescu-sinch: Hi, @JPPortier!
I removed the default behavior of `requireAuthentication` above, and now the controller is the only place where the developer specifies the authentication requirement. That's the single source of truth.
What do you think?

examples/templates/webhooks/appsettings.json
rpredescu-sinch: Hi, @JPPortier!

It's not strictly required because by default, ASP.NET Core allows all hosts ("*") when not specified.
For our webhook example, keeping `"AllowedHosts": "*"` has the advantage of instructing the developers that they need to restrict the `AlloweHosts` to Sinch domain in production.

I have added the following documentation in README.md:

### AllowedHosts configuration
The `appsettings.json` sets `"AllowedHosts": "*"` to allow all hosts during development, which is necessary for webhook development where external services (e.g., nGrok) must be able to POST (forward requests) to your endpoint. **In production, you should restrict this to Sinch's domain(s) for security.**

What do you think?

examples/templates/webhooks/README.md
rpredescu-sinch: Hi, @JPPortier!
Now using the markdown links to `appsettings.json` where configuration must be set.
Thank you!

examples/templates/webhooks/README.md
rpredescu-sinch: Hi, @JPPortier!
Added a markdown link to `Properties/launchSettings.json`.
Thank you!

examples/templates/webhooks/README.md
rpredescu-sinch: Hi, @JPPortier!
Good catch, that was a mismatch.
I simplified the documentation to instruct users to navigate to the webhooks folder first, after which they can execute simplified commands:
```
dotnet restore
dotnet run
```
This removes ambiguity about the working directory.

examples/templates/webhooks/README.md
rpredescu-sinch: Hi, @JPPortier!
I updated README.md to provide a markdown link to the files to be edited.
Thank you!

examples/templates/webhooks/Program.cs
rpredescu-sinch: Hi, @asein-sinch!
I added a note in `Program.cs` clarifying that unified credentials aren’t required for webhook validation, and are only needed if users want to respond via the SMS API.
Thank you!

examples/templates/webhooks/appsettings.json
JPPortier: > It's not strictly required because by default, ASP.NET Core allows all hosts ("*") when not specified.

If not required then do not need to be present onto settings file

A note could be added onto README about best practice but just to highlight it. Purpose of skeleton is not to deliver a detailed guide about how to setup a public backend server: there a lot of dedicated material onto internet to serve this purpse.

examples/templates/webhooks/README.md
JPPortier: The README text say `ensureValidAuthentication = false` but WebhooksController.cs file is based onto `requireAuthentication: false` usage

src/Sinch/SMS/Hooks/TextMessage.cs
rpredescu-sinch: Hi, @asein-sinch!

After our discussion I introduced the following refactoring:
- Consolidated duplicate webhook models (`TextMessage` -> `SmsInbound`, etc.)
- Added missing `JsonPropertyName` attributes
- Moved delivery reports to `DeliveryReports` namespace
- Created more granular interfaces like `IBatchDeliveryReport`, `IRecipientDeliveryReport`
- Moved `ISmsEvent` and `SmsEventConverter` to `Sinch.SMS` root because they cover both `DeliveryReports` and `Inbounds`

Thank you for the feedback!

src/Sinch/SMS/Inbounds/SmsInbound.cs
```diff
@@ -7,55 +9,81 @@ public sealed class SmsInbound : IInbound
...
-
+        [JsonPropertyName("id")]
         public required string Id { get; set; }
 
-
+        /// <summary>
+        ///     Gets or sets the Type.
+        /// </summary>
+        [JsonPropertyName("type")]
+        public SmsType Type { get; set; } = SmsType.Text;
```
asein-sinch: Does it make sense to expose `set` ? What about `internal set` ?

src/Sinch/SMS/ISmsEvent.cs
```diff
@@ -0,0 +1,13 @@
+using System.Text.Json.Serialization;
+
+namespace Sinch.SMS
+{
+    /// <summary>
+    ///     Base interface for all SMS events (inbound messages and delivery reports).
+    ///     Used by webhooks and REST API deserialization to route events to concrete types.
+    /// </summary>
+    [JsonConverter(typeof(SmsEventConverter))]
+    public interface ISmsEvent
```
asein-sinch, line 10: Since we are removing the support of .NET 6 in this new SDK version, I'm wondering if it would make sense to use .NET 7+ feature such as `JsonPolymorphic` and `JsonDerivedType`, and changing this interface into an abstract class instead. WDYT?

src/Sinch/SMS/SmsEventConverter.cs
```diff
@@ -0,0 +1,94 @@
...
+                // Batch delivery reports
+                "delivery_report_sms" => element.Deserialize<BatchDeliveryReportSms>(options),
+                "delivery_report_mms" => element.Deserialize<BatchDeliveryReportMms>(options),
+
+                // Recipient delivery reports
+                "recipient_delivery_report_sms" => element.Deserialize<RecipientDeliveryReportSms>(options),
+                "recipient_delivery_report_mms" => element.Deserialize<RecipientDeliveryReportMms>(options),
+
+                // Unknown type
+                _ => null
```
asein-sinch, line 58: Maybe add use a `UnknownSmsEvent` instead to avoid null reference exceptions later

examples/templates/webhooks/appsettings.json
rpredescu-sinch: Hi, @JPPortier!
Removed unnecessary configuration.
Best!

examples/templates/webhooks/README.md
rpredescu-sinch: Good catch!
Replaced the wrong `ensureValidAuthentication` parameter name with the correct one `requireAuthentication`.
Thank you!

src/Sinch/SMS/Hooks/SmsWebhooks.cs
rpredescu-sinch: Hi, @asein-sinch!
Removed unused member variable.
Thank you!

examples/templates/webhooks/README.md
```diff
@@ -0,0 +1,70 @@
...
+- [`Sinch:ProjectId`](appsettings.json) - your Sinch project id
+- [`Sinch:KeyId`](appsettings.json) - your Sinch API key id
+- [`Sinch:KeySecret`](appsettings.json) - your Sinch API key secret
+- [`Sinch:Sms:WebhookSecret`](appsettings.json) - optional shared secret used to validate incoming webhook signatures
+
+### AllowedHosts configuration
+The `appsettings.json` sets `"AllowedHosts": "*"` to allow all hosts during development, which is necessary for webhook development where external services (e.g., nGrok) must be able to POST (forward requests) to your endpoint. **In production, you should restrict this to Sinch's domain(s) for security.**
+
+### Server URLs and ports
+- Development launch settings ([Properties/launchSettings.json](Properties/launchSettings.json)) set the local URLs to:
```
JPPortier: There is no "Properties" directory nor "launchSettings.json" file

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportSms.cs
```diff
@@ -22,7 +24,7 @@ public sealed class DeliveryReport
...
```
JPPortier: There is CI warnings onto this file

src/Sinch/SMS/DeliveryReports/RecipientDeliveryReportSms.cs
```diff
@@ -42,7 +44,7 @@ public sealed class RecipientDeliveryReport
         ///     The recipient delivery report type.
         /// </summary>
         [JsonPropertyName("type")]
-        public RecipientDeliveryReportType? Type { get; set; }
+        public RecipientDeliveryReportType? Type { get; set; } = RecipientDeliveryReportType.Sms;
```
JPPortier: There is CI warnings onto this file

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportSms.cs
rpredescu-sinch: Hi, @asein-sinch !
I have removed the setter and expose only a getter, so the property is always `Sms` and cannot be changed. I applied the same pattern for the other classes implementing `Type`.
Thank you!

src/Sinch/SMS/Hooks/BinaryMessage.cs
rpredescu-sinch: Hi, @asein-sinch!
Removed `BinaryMessage` class and now using original `BinaryInbound` where `SentAt` is optional.
Thank you!

src/Sinch/SMS/Hooks/BinaryMessage.cs
rpredescu-sinch: Hi, @asein-sinch!
Good catch! I removed `BinaryMessage` as part of a refactoring and now using `BinaryInbound` which has a default type (enforced by removing the setter):
`public SmsType Type => SmsType.Binary;`
Thank you!

src/Sinch/SMS/Hooks/BinaryMessage.cs
rpredescu-sinch: Hi, @asein-sinch!
I removed the `BinaryMessage` class in favor of the existing `BinaryInbound`, and added the missing `ToString()` implementation. There was no reason not to have the string representation of the model, I just missed it in the first place.
Thank you!

src/Sinch/SMS/Hooks/MediaMessage.cs
rpredescu-sinch: Hi, @asein-sinch!
I removed `MediaMessage` as part of a refactoring, and now using `MediaInbound` where we enforce a default type by removing the setter:
`public SmsType Type => SmsType.Media;`
Thank you!

src/Sinch/SMS/Hooks/MediaMessageBodyDetails.cs
rpredescu-sinch: Hi @asein-sinch!
I removed `MediaMessageBodyDetails` in favor of original `MmsMedia` which has a required Status property.
Thank you!

src/Sinch/SMS/DeliveryReports/RecipientDeliveryReportMms.cs
rpredescu-sinch: Hi, @asein-sinch!
In the end I have introduced `DeliveryReceiptStatusCode` enum for the status code to match the API schema:
https://developers.sinch.com/docs/sms/api-reference/sms/delivery-reports/delivery-report-error-codes#delivery-reports/Delivery-report-error-codes
What do you think?

tests/Sinch.Tests/Sms/WebhooksTests.cs
rpredescu-sinch: Hi, @asein-sinch!
I removed `MediaMessage` and `BinaryMessage` and I'm now relying on previous `BinaryInbound`, `MediaInbound` and `SmsInbound` classes which have the type defined in the class as immutable property.
I have update the webhook tests and removed the forbidden now assignment: `Type = Sinch.SMS.Inbounds.SmsType.Text`
Let me know what do you think.
Thank you!

tests/Sinch.Tests/Sms/WebhooksTests.cs
rpredescu-sinch: Good catch, @asein-sinch!
I refactored the `Type` property to an `enum` as you noticed it should have been from the start.
See my comment above: https://github.com/sinch/sinch-sdk-dotnet/pull/153#discussion_r2754233767
Thank you!

tests/Sinch.Tests/Resources/Sms/Inbounds/InboundText.json
rpredescu-sinch: Good catch, @asein-sinch!
The test payloads in `Inbounds` folder are not used and I deleted them. Here is the more comprehensive versions which are actually used from the `Hooks` folder:
- tests\Sinch.Tests\Resources\Sms\Hooks\InboundText.json
- tests\Sinch.Tests\Resources\Sms\Hooks\InboundMedia.json
- tests\Sinch.Tests\Resources\Sms\Hooks\InboundBinary.json

Thank you!

tests/Sinch.Tests/Resources/Sms/Inbounds/InboundBinary.json
rpredescu-sinch: Hi, @asein-sinch!
Indeed, the test files in Inbounds folder are not used and I deleted them.
See my comment above: https://github.com/sinch/sinch-sdk-dotnet/pull/153#discussion_r2754316630
Thank you!

src/Sinch/SMS/Inbounds/SmsInbound.cs
rpredescu-sinch: Hi, @asein-sinch!
You are right and no setter is needed. 
I have refactored the `Type` property across the classes implementing `IInbound` and now we only have a getter and default value:
`public SmsType Type => SmsType.Text;`
What do you think?
Thank you!

examples/templates/webhooks/README.md
rpredescu-sinch: Hi @JPPortier!
I removed the section entirely because it contained standard .NET configuration details which are not needed for .NET developers. They can configure it locally so no need to push the properties folder to the repository in the end.
The default ports are already mentioned in the **## Use ngrok to forward request to local server** section, so we're covered there.
What do you think?
Thank you!

JPPortier: If part of "standard C# framework" and well known be developers ->  👌

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportSms.cs
rpredescu-sinch: Hi @JPPortier!
I have refactored the `Type` definition and it is now:
`public DeliveryReportType Type => DeliveryReportType.Sms;`
which should solve the CI issue.
Thank you!

src/Sinch/SMS/DeliveryReports/RecipientDeliveryReportSms.cs
rpredescu-sinch: Hi @JPPortier!
See my comment above: https://github.com/sinch/sinch-sdk-dotnet/pull/153#discussion_r2754817612
Thank you!

src/Sinch/SMS/SmsEventConverter.cs
rpredescu-sinch: Hi, @asein-sinch!
This is a good subject for a future discussion, but I think introducing an unknown type will trigger a chain of changes which will make the PR exceed its scope. I'll schedule a chat about it instead, what do you think?
Best!

src/Sinch/SMS/ISmsEvent.cs
rpredescu-sinch: Hi, @asein-sinch!
This would shape how we handle events going forward, so I’d like to investigate and discuss it before we introduce the change.
What do you think?

asein-sinch: Sure, let's create a separate task for this to assess the impact. I'll let you resolve the conversation once the ticket is created.

src/Sinch/SMS/SmsEventConverter.cs
asein-sinch: After discussion, the implementation is fine as it's caught above to throw an `InvalidOperationException`. Maybe this is not the best exception type to throw. Let's review that in the other ticket about the `SinchApiException`

src/Sinch/SMS/Inbounds/MediaInbound.cs
```diff
@@ -11,13 +11,18 @@ namespace Sinch.SMS.Inbounds
     /// </summary>
     public sealed class MediaInbound : IInbound
     {
+        /// <summary>
+        ///     Gets or sets the Type.
+        /// </summary>
+        [JsonPropertyName("type")]
+        public SmsType Type => SmsType.Media;
```
asein-sinch: I'd rename `SmsType` into `MessageType`, `InboundType` or `InboundMessageType` just to not have the confusion SMS <=> MMS

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportMms.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// The delivery report type.
+        /// </summary>
+        [JsonPropertyName("type")]
+        public DeliveryReportType Type => DeliveryReportType.Mms;
+
+        /// <summary>
+        /// Batch ID
+        /// </summary>
+        [JsonPropertyName("batch_id")]
+        public string BatchId { get; set; } = string.Empty;
```
asein-sinch: Would it make sense to set the property `required` instead and not use a default empty value?

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportMms.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// Client reference, if any
+        /// </summary>
+        [JsonPropertyName("client_reference")]
+        public string? ClientReference { get; set; }
+
+        /// <summary>
+        /// Statuses for the batch
+        /// </summary>
+        [JsonPropertyName("statuses")]
+        public List<DeliveryReportStatus>? Statuses { get; set; }
```
asein-sinch: Required property

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportSms.cs
asein-sinch: Required property

asein-sinch: Required property

src/Sinch/SMS/Inbounds/MediaInbound.cs
rpredescu-sinch: Done, thank you!

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportMms.cs
rpredescu-sinch: Hi, @asein-sinch!
Definitely!
Thank you!

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportMms.cs
rpredescu-sinch: Done, thank you!

src/Sinch/SMS/DeliveryReports/BatchDeliveryReportSms.cs
rpredescu-sinch: Done, thank you!

rpredescu-sinch: Done, thank you!

asein-sinch (approved):

PR #155, opened 2026-02-05
Conversation apps e2e tests

rpredescu-sinch:
### Summary
Conversation Apps e2e tests.

### Details
.NET SDK running without failure supporting Conversation Apps e2e tests.


src/Sinch/SinchClient.cs
```diff
@@ -188,9 +188,12 @@ public SinchClient(SinchClientConfiguration clientConfiguration)
...
                 var conversationConfig = _sinchClientConfiguration.ConversationConfiguration;
-                var conversationBaseAddress =
-                    conversationConfig.ResolveConversationUrl();
-                var templatesBaseAddress = conversationConfig.ResolveTemplateUrl();
+                var conversationBaseAddress = ResolveUrl(
+                    _sinchClientConfiguration.SinchOptions?.ApiUrlOverrides?.ConversationUrl,
+                    conversationConfig.ResolveConversationUrl);
+                var templatesBaseAddress = ResolveUrl(
+                    _sinchClientConfiguration.SinchOptions?.ApiUrlOverrides?.TemplatesUrl,
+                    conversationConfig.ResolveTemplateUrl);
```
asein-sinch, line 196: Note: this will have to be refactored to use the new configuration for V2
 - at the root of the configuration (ie same level as the projectId) will be the region and the override URL
 - if the override URL is set => it will be used as is
 - otherwise the templated production URL (ending with api.sinch.com) will be resolved with the region
 - if no override URL and no region are defined => raise a error

tests/Sinch.Tests.Features/Conversation/Apps.cs
```diff
@@ -0,0 +1,127 @@
...
+
+    [When(@"I send a request to create an app")]
+    public async Task WhenISendARequestToCreateAnApp()
+    {
+        _app = await _apps.Create(new CreateAppRequest
+        {
+            DisplayName = "E2E Conversation App",
+            ChannelCredentials =
+            [
+                new ConversationChannelCredentials(new StaticBearerCredentials
```
JPPortier: Not coming from this PR but some credentials related classes are within `Apps/Credentials` and some are in `Apps/`

rpredescu-sinch: Hi, @JPPortier!
Moved `ConversationChannelCredentials`, `InstagramCredentials` and `LineEnterpriseCredentials` to new `namespace Sinch.Conversation.Apps.Credentials` and updated MIGRATION_GUIDE.md.
Thank you!

JPPortier: There is still https://github.com/sinch/sinch-sdk-dotnet/blob/feature/DEVEXP-1253-conversation-apps-e2e/src/Sinch/Conversation/Apps/AppleBusinessChatCredentials.cs

JPPortier: Could you check how to provide an helper function when configuring an App with proper credentials ?

With current example it highlight you need to know Sms channel is requiring a StaticBearerCredential.

Ideally; the SDK should provide guidance about channel -> required credential schema.

e.g:
- https://github.com/sinch/sinch-sdk-java/blob/feat/V2.0-next/client/src/main/com/sinch/sdk/domains/conversation/models/v1/credentials/ConversationChannelCredentialsBuilderFactory.java
- https://github.com/sinch/sinch-sdk-node/blob/v1.4-next/packages/conversation/src/models/v1/conversation-channel-credential/conversation-channel-credential.ts
-

tests/Sinch.Tests.Features/Conversation/Utils.cs
JPPortier, line 1: Highlighted this morning, this Utils could be common to all e2e and not having a dedicated onto for Sms, another Conversation, Numbers,...

src/Sinch/SinchClient.cs
JPPortier, line 196: Why is it required now for e2e ?

There was an issue ?

tests/Sinch.Tests.Features/Conversation/Utils.cs
asein-sinch: I'm not sure about that, it's preferable to keep things tight to their scope and create the SinchClient in the "given" step. For Apps for instance, there is no need to initialize the SinchClient with the ConversationTemplate URL

src/Sinch/SinchClient.cs
rpredescu-sinch: Hi, @JPPortier!
Yes, without the URL override, the E2E tests will try to access real API and fail.
Thank you!

rpredescu-sinch: Hi, @asein-sinch!
Does it makes sense to create a separate ticket for this refactoring because the amount of changes could be significant?
Thank you!

tests/Sinch.Tests.Features/Conversation/Apps.cs
rpredescu-sinch: Hi, @JPPortier!
I've added a `ConversationChannelCredentialsBuilderFactory` following the same pattern as the Java SDK. 

The factory provides static methods for each channel (`WhatsApp()`, `Sms()`, `Line()`, etc.) that automatically set the correct channel type for the given credentials. This way developers don't need to remember which credential type maps to which channel.

I've also updated the tests to demonstrate this pattern.

Thank you!

src/Sinch/SinchClient.cs
JPPortier: A cross over I think with comment below.

This mean the sinchclient configuration was not working as expected and detected with these e2e?

I mean we should not have dedicated code in SinchClient to satisfy e2e tests conditions.

asein-sinch: Yes, but please prioritize it before the other Conversation subdomains

rpredescu-sinch: Hi, @JPPortier!
Yes, Sinch client configuration had a bug which prevented E2E to perform properly. We can refactor the relationship between sinch configuration and E2E tests in a separate ticket where we can also handle Antoine's concern below:
https://github.com/sinch/sinch-sdk-dotnet/pull/155#discussion_r2768265479
What do you think?
Thank you!

rpredescu-sinch: Created the ticket!
Thank you!

JPPortier (approved):

PR #156, opened 2026-02-06
Fax services end-to-end tests

rpredescu-sinch:
### Summary

Support end-to-end tests for Fax services.

### DIFF
--- a/src/Sinch/Core/Utils.cs
+++ b/src/Sinch/Core/Utils.cs
@@ -89,6 +89,23 @@ namespace Sinch.Core
             return page * pageSize >= totalCount;
         }
 
+        public static int GetAutoPage(int? page, PageStart pageStart = PageStart.Zero)
+        {
+            switch (pageStart)
+            {
+                case PageStart.Zero:
+                    page ??= 0;
+                    break;
+                case PageStart.One:
+                    page = page is null or < 1 ? 1 : page;
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(pageStart), pageStart, null);
+            }
+
+            return page.Value;
+        }  
+
         public static string ToSnakeCaseQueryString<T>(T obj) where T : class
         {
             var props = typeof(T).GetProperties(BindingFlags.Instance | BindingFlags.Public |
--- a/src/Sinch/Core/Utils.cs
+++ b/src/Sinch/Core/Utils.cs
@@ -89,23 +89,6 @@ namespace Sinch.Core
             return page * pageSize >= totalCount;
         }
 
-        public static int GetAutoPage(int? page, PageStart pageStart = PageStart.Zero)
-        {
-            switch (pageStart)
-            {
-                case PageStart.Zero:
-                    page ??= 0;
-                    break;
-                case PageStart.One:
-                    page = page is null or < 1 ? 1 : page;
-                    break;
-                default:
-                    throw new ArgumentOutOfRangeException(nameof(pageStart), pageStart, null);
-            }
-
-            return page.Value;
-        }  
-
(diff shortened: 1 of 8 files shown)


src/Sinch/Fax/Services/ServicesClient.cs
```diff
@@ -116,25 +116,26 @@ Task<ListNumbersResponse> ListNumbers(string serviceId, int? page, int? pageSize
...
         /// <returns></returns>
-        IAsyncEnumerable<ServicePhoneNumber> ListNumbersAuto(string serviceId, int? page, int? pageSize,
+        IAsyncEnumerable<ServicePhoneNumber> ListNumbersAuto(string serviceId, int? page = null, int? pageSize = null,
             CancellationToken cancellationToken = default);
     }
 
     internal sealed class ServicesClient : ISinchFaxServices
     {
+        private const int DefaultPageSize = 20;
+        private const int EmailsDefaultPageSize = 100;
```
asein-sinch: No, the SDK should not dictate what the API can change at any moment

src/Sinch/Fax/Emails/Emails.cs
asein-sinch, line 1: I was surprised to see that you had to update this class but not the one for Faxes, where there are also paginated responses: the answer is the class `ListFaxResponse` doesn't extend `PagedResponse`: keep this in mind when you'll handle this subdomain

src/Sinch/Fax/PagedResponse.cs
```diff
@@ -1,11 +1,15 @@
...
+
 namespace Sinch.Fax
 {
     public class PagedResponse
     {
         /// <summary>
         ///     Current page
         /// </summary>
-        public int PageNumber { get; set; }
+        /// TODO: Remove [JsonPropertyName("pageNumber")] after mock server update.
```
asein-sinch: Just go with the right implementation and relaunch the build when the mockserver is merged (I've created the pull request already)

rpredescu-sinch: Hi, @asein-sinch!
Removed temporary attribute decoration and now we match the API page property.
Thank you!

src/Sinch/Core/Utils.cs
```diff
@@ -89,6 +89,11 @@ public static bool IsLastPage(int page, int pageSize, int totalCount, PageStart
             return page * pageSize >= totalCount;
         }
 
+        public static bool IsNotLastPage(int page, int pageSize, int totalCount, PageStart pageStart = PageStart.Zero)
+        {
+            return !IsLastPage(page, pageSize, totalCount, pageStart);
+        }
```
asein-sinch, line 95: Not for this PR: this implementation is not OK because it's part of the Core package, making think it's generic, but it's not, depending on whether the API pagination starts with 0 or 1. Why would the API starting with 0 be preferred over the other one? I think the `pageStart` parameter should not be defaulted and be required by all the API for being more explicit.

src/Sinch/Fax/Services/ServicesClient.cs
```diff
@@ -60,7 +60,7 @@ public interface ISinchFaxServices
         /// <param name="pageSize"></param>
         /// <param name="cancellationToken"></param>
         /// <returns></returns>
-        Task<ListServicesResponse> List(int? page, int? pageSize, CancellationToken cancellationToken = default);
+        Task<ListServicesResponse> List(int? page = null, int? pageSize = null, CancellationToken cancellationToken = default);
```
asein-sinch, line 63: Thank you for improving the developer experience 👍

asein-sinch (approved):

PR #157, opened 2026-02-10
Add missing fax services code snippets

rpredescu-sinch:
### Summary

Add missing fax services code snippets.

### Related to
- https://github.com/sinch/sinch-sdk-dotnet/pull/156

### Note
Will target v2-next after [PR-156](https://github.com/sinch/sinch-sdk-dotnet/pull/156) is merged.


examples/snippets/fax/emails/ListEmailsForNumber/ListEmailsForNumber.cs
```diff
@@ -0,0 +1,30 @@
...
+        KeySecret = ConfigurationHelper.GetKeySecret() ?? "MY_KEY_SECRET"
+    }
+});
+
+const string serviceId = "my-service-id";
+const string phoneNumber = "my-virtual-number";
+
+Console.WriteLine($"Listing emails for phone number {phoneNumber} in service {serviceId}");
+
+var response = await sinchClient.Fax.Services.ListEmailsForNumber(serviceId, phoneNumber, page: 1, pageSize: 10);
```
asein-sinch: I think this operation moved to the subdomain `emails` only (even if technically is uses the `services` path)

examples/snippets/fax/services/Delete/Delete.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    SinchUnifiedCredentials = new SinchUnifiedCredentials
+    {
+        ProjectId = ConfigurationHelper.GetProjectId() ?? "MY_PROJECT_ID",
+        KeyId = ConfigurationHelper.GetKeyId() ?? "MY_KEY_ID",
+        KeySecret = ConfigurationHelper.GetKeySecret() ?? "MY_KEY_SECRET"
+    }
+});
+
+const string serviceId = "my-service-id";
```
asein-sinch:
```suggestion
// The Fax Service ID you want to delete
const string serviceId = "FAX_SERVICE_ID";
```

examples/snippets/fax/services/ListAuto/ListAuto.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        ProjectId = ConfigurationHelper.GetProjectId() ?? "MY_PROJECT_ID",
+        KeyId = ConfigurationHelper.GetKeyId() ?? "MY_KEY_ID",
+        KeySecret = ConfigurationHelper.GetKeySecret() ?? "MY_KEY_SECRET"
+    }
+});
+
+Console.WriteLine("Listing all fax services");
+
+await foreach (var service in sinchClient.Fax.Services.ListAuto())
```
asein-sinch, line 26: Design question: for `List` operation, it makes sense to define the page and pageSize parameters, but for `ListAuto`, does it make sense to propose the same parameters?

examples/snippets/fax/services/Create/Create.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        ProjectId = ConfigurationHelper.GetProjectId() ?? "MY_PROJECT_ID",
+        KeyId = ConfigurationHelper.GetKeyId() ?? "MY_KEY_ID",
+        KeySecret = ConfigurationHelper.GetKeySecret() ?? "MY_KEY_SECRET"
+    }
+});
+
+Console.WriteLine("Creating a new fax service");
+
+var request = new CreateServiceRequest
```
asein-sinch: With the arrival of new APIs and the standardisation of the "service" concept, we may find the same class name in different namespaces. I'd suggest to rename `CreateServiceRequest` into `CreateFaxServiceRequest`

examples/snippets/fax/emails/ListEmailsForNumber/ListEmailsForNumber.cs
rpredescu-sinch: Hi, @asein-sinch!
Moved the two list emails for number snippets to `fax/emails` folder.
Thank you!

examples/snippets/fax/services/Create/Create.cs
rpredescu-sinch: @asein-sinch 
Doing the same for the following two FAX services models.
- `UpdateServiceRequest` -> `UpdateFaxServiceRequest`
- `ListServicesResponse` ->  `ListFaxServicesResponse`
- `ListServiceNumbersResponse` -> `ListFaxServiceNumbersResponse`

examples/snippets/fax/services/ListAuto/ListAuto.cs
rpredescu-sinch: Hi @asein-sinch!
I think the snippets are more descriptive if we add `page: 1, pageSize: 20` to ListAuto calls:
- users learn what pagination parameters are available even when using the auto pagination methods
- both `List` and `ListAuto` snippets demonstrate the same capabilities

I have also added this comment:
`// ListAuto handles pagination automatically, but you can control the page size`

What do you think?

examples/snippets/fax/services/Delete/Delete.cs
rpredescu-sinch: Hi @asein-sinch!
I have applied the same pattern to all FAX Services snippets.
Thank you!

examples/snippets/fax/services/ListAuto/ListAuto.cs
```diff
@@ -0,0 +1,29 @@
...
+        ProjectId = ConfigurationHelper.GetProjectId() ?? "MY_PROJECT_ID",
+        KeyId = ConfigurationHelper.GetKeyId() ?? "MY_KEY_ID",
+        KeySecret = ConfigurationHelper.GetKeySecret() ?? "MY_KEY_SECRET"
+    }
+});
+
+Console.WriteLine("Listing all fax services");
+
+// ListAuto handles pagination automatically, but you can control the page size
+await foreach (var service in sinchClient.Fax.Services.ListAuto(page: 1, pageSize: 20))
```
JPPortier: Take care about this: by experience snippets will be used and may be not changed regarding parameters.

So when a user will send request the same parameters will be in use (page & pageSize) instead of the default ones managed by backend.

Now with AI consuming snippets to generate sources it could be better to remove un-necessary conditions which will be propagated 'blindly'

examples/snippets/fax/emails/ListEmailsForNumber/ListEmailsForNumber.cs
```diff
@@ -0,0 +1,31 @@
...
+    }
+});
+
+// The Fax Service ID and phone number you want to list emails for
+const string serviceId = "FAX_SERVICE_ID";
+const string phoneNumber = "my-virtual-number";
+
+Console.WriteLine($"Listing emails for phone number {phoneNumber} in service {serviceId}");
+
+var response = await sinchClient.Fax.Services.ListEmailsForNumber(serviceId, phoneNumber, page: 1, pageSize: 20);
```
JPPortier: Take care about this: by experience snippets will be used and may be not changed regarding parameters.

So when a user will send request the same parameters will be in use (page & pageSize) instead of the default ones managed by backend.

Now with AI consuming snippets to generate sources it could be better to remove un-necessary conditions which will be propagated 'blindly'

examples/snippets/fax/emails/ListEmailsForNumberAuto/ListEmailsForNumberAuto.cs
```diff
@@ -0,0 +1,32 @@
...
+});
+
+// The Fax Service ID and phone number you want to list emails for
+const string serviceId = "FAX_SERVICE_ID";
+const string phoneNumber = "my-virtual-number";
+
+Console.WriteLine($"Listing all emails for phone number {phoneNumber} in service {serviceId}");
+
+// ListEmailsForNumberAuto handles pagination automatically, but you can control the page size
+await foreach (var email in sinchClient.Fax.Services.ListEmailsForNumberAuto(serviceId, phoneNumber, page: 1, pageSize: 20))
```
JPPortier: Take care about this: by experience snippets will be used and may be not changed regarding parameters.

So when a user will send request the same parameters will be in use (page & pageSize) instead of the default ones managed by backend.

Now with AI consuming snippets to generate sources it could be better to remove un-necessary conditions which will be propagated 'blindly'

examples/snippets/fax/services/List/List.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        ProjectId = ConfigurationHelper.GetProjectId() ?? "MY_PROJECT_ID",
+        KeyId = ConfigurationHelper.GetKeyId() ?? "MY_KEY_ID",
+        KeySecret = ConfigurationHelper.GetKeySecret() ?? "MY_KEY_SECRET"
+    }
+});
+
+Console.WriteLine("Listing fax services");
+
+var response = await sinchClient.Fax.Services.List(page: 1, pageSize: 20);
```
JPPortier: Take care about this: by experience snippets will be used and may be not changed regarding parameters.

So when a user will send request the same parameters will be in use (page & pageSize) instead of the default ones managed by backend.

Now with AI consuming snippets to generate sources it could be better to remove un-necessary conditions which will be propagated 'blindly'

examples/snippets/fax/services/ListNumbers/ListNumbers.cs
```diff
@@ -0,0 +1,30 @@
...
+        KeySecret = ConfigurationHelper.GetKeySecret() ?? "MY_KEY_SECRET"
+    }
+});
+
+// The Fax Service ID you want to list numbers for
+const string serviceId = "FAX_SERVICE_ID";
+
+Console.WriteLine($"Listing numbers for fax service: {serviceId}");
+
+var response = await sinchClient.Fax.Services.ListNumbers(serviceId, page: 1, pageSize: 20);
```
JPPortier: Take care about this: by experience snippets will be used and may be not changed regarding parameters.

So when a user will send request the same parameters will be in use (page & pageSize) instead of the default ones managed by backend.

Now with AI consuming snippets to generate sources it could be better to remove un-necessary conditions which will be propagated 'blindly'

examples/snippets/fax/services/ListNumbersAuto/ListNumbersAuto.cs
```diff
@@ -0,0 +1,32 @@
...
+    }
+});
+
+// The Fax Service ID for which you want to list numbers for
+const string serviceId = "FAX_SERVICE_ID";
+
+Console.WriteLine($"Listing all numbers for fax service: {serviceId}");
+
+// ListNumbersAuto handles pagination automatically, but you can control the page size
+await foreach (var number in sinchClient.Fax.Services.ListNumbersAuto(serviceId, page: 1, pageSize: 20))
```
JPPortier: Take care about this: by experience snippets will be used and may be not changed regarding parameters.

So when a user will send request the same parameters will be in use (page & pageSize) instead of the default ones managed by backend.

Now with AI consuming snippets to generate sources it could be better to remove un-necessary conditions which will be propagated 'blindly'

examples/snippets/fax/services/ListAuto/ListAuto.cs
asein-sinch: I mean for the `ListAuto`, you are going to iterate on all the pages, so the pageSize parameter will be adjusting the number of pages to fetch (ie number of network transports) versus the quantity of elements to fetch for each transport.
On the opposite, with the `List`, it may be used in the context of a UI where users can fetch a specific page and/or adjust the number of elements displayed on a web page.

Regarding your points, I agree with the first one: it's indeed great to show in the snippets which parameters exist. But for the second point, as I wrote above, I think the 2 methods are serving different use cases so it's not a requirement to demo the same parameters for both

rpredescu-sinch: Hi, @asein-sinch!

Thank you for the detailed feedback!

Your point about different use cases makes sense! `List()` is for specific page fetching (UI), while `ListAuto()` is for full iteration and they do serve different purposes indeed. `ListAuto()`'s API default pagination is probably the right choice for most of the use cases, and it helps us to keep the snippets simple and straight forward. 

I'm now thinking to remove the pagination params not only for `ListAuto()`, but for the `List()` as well because @JPPortier also raises a valid concern below: snippets will most likely get copied as-is by either developers or AI. Hardcoded `page: 1, pageSize: 20` parameters can become production defaults without developers realizing the implications. So better to rely on API defaults if users do not have strong reasons to do otherwise. If they need more pagination control, they can check the method signatures and documentation.

We can revisit this if the we think we're missing something, but lets move forward by removing `page` and `pageSize` parameters from both `List()` and `ListAuto()` snippets?

What do you think?

examples/snippets/fax/services/ListAuto/ListAuto.cs
rpredescu-sinch: Hi @JPPortier!

Thanks for flagging this! 

Better to let developers rely on API defaults and check docs if they need custom pagination.

See my answer to @asein-sinch above, but the main point is that we should move forward with removing `page` and `pageSize` from both` List()` and `ListAuto()` snippets as you suggested (I already pushed the change).

Thank you!

examples/snippets/fax/emails/ListEmailsForNumber/ListEmailsForNumber.cs
rpredescu-sinch: I have answered above: https://github.com/sinch/sinch-sdk-dotnet/pull/157#discussion_r2797980270

JPPortier (approved):

PR #166, opened 2026-02-27
Api error response deserialization issue

rpredescu-sinch:
### Summary

Client identified a bug in the official Sinch .NET SDK (v1.4.1) related to SMS API error handling.

### Problem

The SDK expects an integer for an error code, but the SMS API returns a string. This causes a `JsonException` that breaks all try/catch error handling for the SMS client and loses the original error details.

```
JsonException: The JSON value could not be converted to System.Nullable`1[System.Int32]. Path: $.code | LineNumber: 0 | BytePositionInLine: 41.
```


tests/Sinch.Tests/Core/HttpTests.cs
```diff
@@ -321,6 +321,34 @@ public async Task NewTokenIsFetchedBetweenTwoRequestsStartingFromExpired()
             _httpMessageHandlerMock.VerifyNoOutstandingExpectation();
         }
 
+        [Fact]
+        public async Task SmsApiStringErrorCodeThrowsSinchApiException()
+        {
+            // The SMS API returns { "code": "...", "text": "..." } where code is a string, not an int.
+            // When code is a non-numeric string (e.g. "Forbidden"), JsonNumberHandling.AllowReadingFromString
+            // (set by JsonSerializerDefaults.Web) cannot coerce it to int?, causing a JsonException that
+            // escapes all catch (SinchApiException) blocks. The fix uses IntOrStringConverter on Code.
```
JPPortier: We can remove this comment (there is no longer IntOrStringConverter) and the test name should help to understand the purpose.

If there is tricky point to be document: better to document where is have sense: here it could be within core/Extensions.cs but I have the feeling the code is self explained

tests/Sinch.Tests/Core/HttpTests.cs
```diff
@@ -321,6 +321,34 @@ public async Task NewTokenIsFetchedBetweenTwoRequestsStartingFromExpired()
             _httpMessageHandlerMock.VerifyNoOutstandingExpectation();
         }
 
+        [Fact]
+        public async Task SmsApiStringErrorCodeThrowsSinchApiException()
```
JPPortier, line 325: Do we have a test to ensure "normal" use case: An ApiErrorResponse with integer is also correctly reported

tests/Sinch.Tests/Core/HttpTests.cs
rpredescu-sinch: Hi @JPPortier!
I removed the unnecessary comment.
Best!

tests/Sinch.Tests/Core/HttpTests.cs
rpredescu-sinch: Hi @JPPortier!
I added an unit test covering integer code path.
Thank you!

JPPortier (approved):

rpredescu-sinch:
To be addressed in this PR:
- https://github.com/sinch/sinch-sdk-dotnet/pull/167

PR #171, opened 2026-03-03
Snippets compilation issues and CI

rpredescu-sinch:
### Summary

Snippets solution is missing from CI build and it also has some other issues like incorrect projects path case, unnecessary Sinch NuGet package reference and outdated snippets structure.

### Solution

1. Add Snippets solution to `build.yaml`
2. Update uppercase `fax\Faxes` to lowercase `fax\faxes` in Snippets solution to avoid issues on Linux
3. Remove unnecessary Sinch NuGet package reference


examples/snippets/Snippets.sln
JPPortier, line 1: As a NOT .NET developer: dummy question; there is no way to compile all sub projects without having to list them explicitly ?

Asking, because we can imagine in the future to add a new snippet but missind to update this file, then CI won't complain if any error from snippet

examples/snippets/conversation/capability/LookupByContactId/LookupByContactId.csproj
```diff
@@ -7,10 +7,6 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(Configuration)'!='Debug'">
-    <PackageReference Include="Sinch" />
-  </ItemGroup>
-
   <ItemGroup>
     <ProjectReference Include="..\..\..\..\..\src\Sinch\Sinch.csproj" />
     <ProjectReference Include="..\..\..\Shared\SnippetsConfig.Shared.csproj" />
```
JPPortier, line 12: Some of snippets project files contains ` <ProjectReference Include="..\..\..\Shared\SnippetsConfig.Shared.csproj" />` and some other not.

Is it expected ?

JPPortier (commented):
PR's description say `Update snippets structure with variables first, configuration second and API execution third` but not seeing this within PR

examples/snippets/Snippets.sln
rpredescu-sinch: Hi @JPPortier!

The current setup optimizes for **IDE** usability. The solution gives the best experience in **Visual Studio** / **JetBrains Rider** and ensures new snippets show up for developers.

The downside is **CI** only builds what's in the solution, so a newly added snippet that isn’t added to the `.sln` could be missed.

If we want extra safety, we can switch **CI** to build all snippet `.csproj` for maximum coverage (we will still keep the solution for **IDE**).

It's a trade‑off between **coverage** and **developer experience**. 

Which direction should we go?

examples/snippets/conversation/capability/LookupByContactId/LookupByContactId.csproj
rpredescu-sinch: Hi @JPPortier!

Good catch!

`ConfigurationHelper` is part of `SnippetsConfig.Shared.csproj`, so snippets that use it need the shared project reference. Some older snippets use `Environment.GetEnvironmentVariable` directly instead, which is why they don't have it.

This is the same approach used by the Java SDK, which has a shared `Settings` utility instead of raw env lookups.

`String projectId = Settings.getProjectId().orElse("MY_PROJECT_ID");`

I think we should standardize on `ConfigurationHelper` across all snippets for consistency.

It's more capable since it supports both `appsettings.json` (we do not use them for the snippets) and **environment variables.** 

I'll can update the remaining snippets that still use `Environment.GetEnvironmentVariable` to use `ConfigurationHelper`.

What do you think?

rpredescu-sinch:
> PR's description say `Update snippets structure with variables first, configuration second and API execution third` but not seeing this within PR

Hi @JPPortier!

I removed this line from the PR. 

I'm thinking that this could be a different PR because of the amount of changes. We already have 57 files changes here, mostly project files.

The snippets structure updates will double the number of changed files (this time .cs files) and will make it difficult to follow which are CI changes and which are snippet formatting changes.

What do you think?

examples/snippets/conversation/capability/LookupByContactId/LookupByContactId.csproj
JPPortier: > I think we should standardize

Yes please

examples/snippets/Snippets.sln
JPPortier: So, to summarize: 
- we can ensure 100% coverage with .csproj from CI
- we can provide proper DX with .sln
- both are possible together

If the answer is 'yes' to all of questions above: we need to go this way

rpredescu-sinch: Hi, @JPPortier!

I have added a `dirs.proj` file uses a wildcard (**/*.csproj) to auto-discover all `.csproj` files under the **Snippets** directory. If someone adds a new snippet project, CI will automatically pick it up and build it.
The Snippets.sln file remains for local development.

The CI build command now looks like this:
```
- name: Build Snippets
        run: dotnet build examples/snippets/dirs.proj
```
Best!

.github/workflows/build.yaml
```diff
@@ -73,6 +73,9 @@ jobs:
       - name: Build
         run: dotnet build -p:Version=${{ steps.gitversion.outputs.semVer }} --configuration Release --no-restore
 
+      - name: Build Snippets
+        run: dotnet build examples/snippets/dirs.proj
+
```
JPPortier, line 78: What do you think about having `dirs.proj` at `examples` level (not under `/snippets`) ?
So it should also validation against the `templates/webhooks` project.

examples/snippets/conversation/capability/LookupByContactId/LookupByContactId.csproj
rpredescu-sinch: Hi @JPPortier!
This will be fixed in https://github.com/sinch/sinch-sdk-dotnet/pull/172.
Best!

.github/workflows/build.yaml
rpredescu-sinch: Hi @JPPortier!

I have moved the `dirs.proj` to examples folder and now its building all the projects there.

Best!

JPPortier (approved):
🚀

PR #177, opened 2026-03-11
OAS Sync Conversation - Payment Order Update

rpredescu-sinch:
### Summary

Sync the .NET SDK Conversation to OAS.

### Changes

- Removes deprecated `OrderDetailsPaymentSettings` 
- Adds `PaymentButtons` with concrete implementations


src/Sinch/Conversation/Messages/Message/ChannelSpecificMessages/WhatsApp/IWhatsAppPaymentButton.cs
```diff
@@ -0,0 +1,20 @@
...
+    ///     <see cref="WhatsAppPaymentSettingsButtonBoleto" />.
+    /// </summary>
+    [JsonConverter(typeof(WhatsAppPaymentButtonJsonConverter))]
+    [JsonDerivedType(typeof(WhatsAppPaymentSettingsButtonPix))]
+    [JsonDerivedType(typeof(WhatsAppPaymentSettingsButtonPaymentLink))]
+    [JsonDerivedType(typeof(WhatsAppPaymentSettingsButtonBoleto))]
+    public interface IWhatsAppPaymentButton
+    {
+        /// <summary>The button type discriminator.</summary>
+        public WhatsAppPaymentButtonType Type { get; }
```
JPPortier: Does it  have to be public ? Information is embedded with inherited classes and is only having interest for serializer/deserializer right ?

src/Sinch/Conversation/Messages/Message/ChannelSpecificMessages/WhatsApp/WhatsAppPaymentButtonJsonConverter.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public override void Write(Utf8JsonWriter writer, IWhatsAppPaymentButton value, JsonSerializerOptions options)
+        {
+            if (value is WhatsAppPaymentSettingsButtonPix pix)
+                JsonSerializer.Serialize(writer, pix, options);
+            else if (value is WhatsAppPaymentSettingsButtonPaymentLink link)
+                JsonSerializer.Serialize(writer, link, options);
+            else if (value is WhatsAppPaymentSettingsButtonBoleto boleto)
+                JsonSerializer.Serialize(writer, boleto, options);
```
JPPortier: Is .NET serializer need a specific method call signature to serialize and object ?
A simple:
```
JsonSerializer.Serialize(writer, value, options);
```
cannot work ?

src/Sinch/Conversation/Messages/Message/ChannelSpecificMessages/WhatsApp/IWhatsAppPaymentButton.cs
rpredescu-sinch: Hi @JPPortier!

In C# interface members are public by definition, so the visibility can't be reduced.

`IWhatsAppPaymentButton` let us assign any mix of `WhatsAppPaymentSettingsButtonPix`, `ButtonPaymentLink`, or `ButtonBoleto` to the same list without needing to know the concrete type. It also simplifies polymorphic deserialization via `WhatsAppPaymentButtonJsonConverter`.

Lets keep the current approach because it is a common C# pattern for handling JSON discriminators.

What do you think?

src/Sinch/Conversation/Messages/Message/ChannelSpecificMessages/WhatsApp/WhatsAppPaymentButtonJsonConverter.cs
rpredescu-sinch: Hi @JPPortier!

Thank you for spotting this!

A great simplification would be:

`JsonSerializer.Serialize(writer, value, value.GetType(), options);` (very close to your suggestion)

Without `value.GetType()`, type would be inferred as `IWhatsAppPaymentButton` at compile time. Since `[JsonConverter(typeof(WhatsAppPaymentButtonJsonConverter))]` is declared on the interface, the serializer would invoke this same Write method again, causing infinite recursion.

Passing `value.GetType()` resolves to the concrete runtime type, which has no converter attached, so serialization proceeds normally (I covered the serialization for the three types with missing unit tests).

I've also applied the same pattern to `ChannelSpecificMessageJsonInterfaceConverter` to avoid the same issue there.

Best!

src/Sinch/Conversation/Messages/Message/ChannelSpecificMessages/WhatsApp/IWhatsAppPaymentButton.cs
rpredescu-sinch: Hi @JPPortier!
I removed Type discriminator from the interface and updated unit tests.
Thank you!

src/Sinch/Conversation/Messages/Message/ChannelSpecificMessages/WhatsApp/WhatsAppPaymentButtonJsonConverter.cs
rpredescu-sinch: Hi @JPPortier!
Updated serialization to `JsonSerializer.Serialize(writer, value, value.GetType(), options); `
`value.GetType()` has nothing to do with the `Type` discriminator and its here to avoid a stack overflow (see the above explanation).
Thank you!

tests/Sinch.Tests/HelpersTests.cs
JPPortier, line 1: ❤️

JPPortier (approved):
🚀

PR #179, opened 2026-03-17
OAS Sync - Add LINE Notification Message Template support

rpredescu-sinch:
### Summary

Implements the NOTIFICATION_MESSAGE_TEMPLATE channel-specific message type for the LINE channel, as defined in the OAS update.

### DIFF
--- a/src/Sinch/Conversation/Messages/Message/AppMessage.cs
+++ b/src/Sinch/Conversation/Messages/Message/AppMessage.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using Sinch.Conversation.Common;
+using Sinch.Conversation.Messages.Message.ChannelSpecificMessages.Line;
 using Sinch.Conversation.Messages.Message.ChannelSpecificMessages.WhatsApp;
 using Sinch.Core;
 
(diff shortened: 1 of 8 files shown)


src/Sinch/Conversation/Messages/Message/AppMessage.cs
```diff
@@ -185,14 +190,37 @@ public sealed class OrderStatusPaymentMessage : IChannelSpecificMessage
...
+        /// </summary>
         public static readonly MessageType OrderDetails = new("ORDER_DETAILS");
+        /// <summary>
+        /// The WhatsApp order status message type.
+        /// </summary>
         public static readonly MessageType OrderStatus = new("ORDER_STATUS");
+        /// <summary>
+        /// LINE notification message template type
+        /// </summary>
+        public static readonly MessageType NotificationMessageTemplate = new("NOTIFICATION_MESSAGE_TEMPLATE");
```
matsk-sinch: Since it's dedicated for Line, consider renaming to `LineNotificationMessageTemplate`

rpredescu-sinch: Hi @matsk-sinch!
Nice catch!
Updated the name.
Thank you!

matsk-sinch (approved):

PR #184, opened 2026-04-06
Conversation messages e2e

rpredescu-sinch:
### Summary

Support end-to-end tests for Conversation messages.

In this PR we have:
- added e2e tests
- added integration tests

### DIFF
--- a/src/Sinch/Conversation/Messages/Messages.cs
+++ b/src/Sinch/Conversation/Messages/Messages.cs
@@ -7,6 +7,7 @@ using System.Threading.Tasks;
 using Sinch.Conversation.Messages.List;
 using Sinch.Conversation.Messages.Message;
 using Sinch.Conversation.Messages.Send;
+using Sinch.Conversation.Messages.Update;
 using Sinch.Core;
 using Sinch.Logger;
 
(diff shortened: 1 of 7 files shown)


src/Sinch/Conversation/Messages/Messages.cs
```diff
@@ -46,6 +47,14 @@ public interface ISinchConversationMessages
         Task<ConversationMessage> Get(string messageId, MessageSource? messagesSource = default,
             CancellationToken cancellationToken = default);
 
+        /// <summary>
+        ///     This operation lists all messages using server-default settings.
+        ///     See <see cref="List(ListMessagesRequest, CancellationToken)" /> to apply filters.
+        /// </summary>
+        /// <param name="cancellationToken">A token to cancel the asynchronous operation.</param>
+        /// <returns>A <see cref="ListMessagesResponse"/> containing the first page of messages.</returns>
+        Task<ListMessagesResponse> List(CancellationToken cancellationToken = default);
```
JPPortier, line 56: Interesting inconsistency with ListMessagesResponse and other classes (e.g.  batches, fax, ...)
For some it is a List but here it is an IEnumerable

tests/Sinch.Tests/Conversation/MessagesTests.cs
JPPortier: The issue here is not the complex object, but we don't know which kind of class/object it have to reflect.

JPPortier (approved):

PR #186, opened 2026-04-08
Conversation Conversations e2e

rpredescu-sinch:
### Summary

Support end-to-end tests for Conversation conversations.

In this PR we have:
- added e2e tests
- added integration tests


src/Sinch/Conversation/Conversations/Create/CreateConversationRequest.cs
```diff
@@ -15,36 +15,32 @@ public sealed class CreateConversationRequest
...
-
+        /// <summary>
+        ///     An arbitrary identifier that can be used to correlate this conversation with resources in external systems.
+        /// </summary>
+        public string? CorrelationId { get; set; }
 
         /// <summary>
         ///     Arbitrary data set by the Conversation API clients. Up to 1024 characters long.
         /// </summary>
         public string? Metadata { get; set; }
```
JPPortier: This field should be tagged as `Obsolete`

src/Sinch/Conversation/Conversations/InjectMessage/InjectMessageRequest.cs
JPPortier: Why ignoring this field ?
It is well defined from specification

JPPortier, line 12: This class is wrongly defining payload: current definition enable to define at same time contact message AND app message. But it should an XOR

src/Sinch/Conversation/Conversations/List/ListConversationsRequest.cs
```diff
@@ -3,10 +3,9 @@ namespace Sinch.Conversation.Conversations.List
     public sealed class ListConversationsRequest
     {
         /// <summary>
-        ///     Required. True if only active conversations should be listed.
+        ///     True if only active conversations should be listed.
```
JPPortier: This is not reflecting the current specification description content

MIGRATION_GUIDE.md
```diff
@@ -865,3 +867,46 @@ var appEvent = conversationEvent.AppEvent;
...
+var request = new ListConversationsRequest { OnlyActive = false, AppId = "app-id" };
+```
+
+Version 2.*:
+```csharp
+// OnlyActive can be omitted — the server uses its default behavior
+var request = new ListConversationsRequest { AppId = "app-id" };
+// Or explicitly set it
+var request = new ListConversationsRequest { OnlyActive = true, AppId = "app-id" };
+```
```
JPPortier: Please remove

src/Sinch/Conversation/Conversations/Create/CreateConversationRequest.cs
rpredescu-sinch: Hi, @JPPortier !
Marked the field as `[Obsolete]`.
Thank you!

src/Sinch/Conversation/Conversations/InjectMessage/InjectMessageRequest.cs
rpredescu-sinch: Hi @JPPortier!
Updated class definition to accept either `AppMessage` or `ContactMessage` at initialization.
Thank you!

rpredescu-sinch: Hi @JPPortier!
Removed the `[JsonIgnore]` to comply with OAS. 
The reason why it was ignore was that it was used in the in the URL path, so the conversation ID comes from the path parameter.
Thank you!

src/Sinch/Conversation/Conversations/List/ListConversationsRequest.cs
rpredescu-sinch: Hi @JPPortier!
Updated inline documentation to follos OAS.
Best!

MIGRATION_GUIDE.md
rpredescu-sinch: Hi, @JPPortier!
Removed unnecessary  example line.
Thank you!

src/Sinch/Conversation/Conversations/Create/CreateConversationRequest.cs
JPPortier: My bad @rpredescu-sinch 

Thinkig a second time: because of we are moving to SDK V2 we are open to breaking changes and should remove the field support from SDK (and so it will remove all compilation warnings about deprecation of field)

I apologize for this second opinion

src/Sinch/Conversation/Conversations/InjectMessage/InjectMessageRequest.cs
JPPortier: > so the conversation ID comes from the path parameter.

Not exactly: the conversation_id is not coming from: it have to be set for path parameter and within payload request to.

src/Sinch/Conversation/Conversations/Create/CreateConversationRequest.cs
rpredescu-sinch: Hi, @JPPortier !
No worries, my bad!
I removed all the `[Obsolete]` tags previously, so I should have remembered that we do not need the annotation for v2. Fixed it.
Thank you!

rpredescu-sinch: Hi @JPPortier !
I removed unused `Metadata` property for V2.
Best!

tests/Sinch.Tests.Features/Conversation/Conversations.cs
```diff
@@ -0,0 +1,315 @@
...
+    [Then(@"the response contains the conversation details")]
+    public void ThenTheResponseContainsTheConversationDetails()
+    {
+        _conversation.Should().NotBeNull();
+        _conversation.Id.Should().Be(ConversationId001);
+        _conversation.AppId.Should().Be(AppId001);
+        _conversation.ContactId.Should().Be(ContactId002);
+        _conversation.ActiveChannel.Should().Be(ConversationChannel.Messenger);
+        _conversation.Active.Should().BeTrue();
+        _conversation.Metadata.Should().Be("e2e tests");
```
JPPortier: This field too is tagged as `deprecated` from specs: https://developers.sinch.com/docs/conversation/api-reference/conversation/conversation/conversation_getconversation#conversation/conversation_getconversation/t=response&c=200&path=metadata

Good timing to remove it support from SDK V2

tests/Sinch.Tests.Features/Conversation/Conversations.cs
```diff
@@ -0,0 +1,315 @@
...
+using Sinch.Conversation.Common;
+using Sinch.Conversation.Conversations;
+using Sinch.Conversation.Conversations.Create;
+using Sinch.Conversation.Conversations.InjectEvent;
+using Sinch.Conversation.Conversations.InjectMessage;
+using Sinch.Conversation.Conversations.List;
+using Sinch.Conversation.Events;
+using Sinch.Conversation.Events.EventTypes;
+using Sinch.Conversation.Messages.Message;
+using ConversationEntity = Sinch.Conversation.Conversations.Conversation;
```
JPPortier: Why renaming `Conversation` with `ConversationEntity` ?

Is it required ?

Same question for https://github.com/sinch/sinch-sdk-dotnet/blob/feature/DEVEXP-1255-conversation-conversations-e2e/tests/Sinch.Tests/Conversation/ConversationsTests.cs#L19

If it is required (😢), please keep a consistent naming convention.

tests/Sinch.Tests/Conversation/ConversationsTests.cs
```diff
@@ -1,17 +1,407 @@
...
+
+        #endregion
+
+        #region Model Tests
+
         [Fact]
         public void UpdateMaskConversation()
         {
-            var conversation = new Sinch.Conversation.Conversations.Conversation
+            var conversation = new ConversationModel
```
JPPortier: Previous implementation seems better: ConversationModel is an unknown class and need to find what it is about.
Seeing `Sinch.Conversation.Conversations.Conversation` seems to be more straightforward when reading (if `Conversation` alone is not possible)

tests/Sinch.Tests.Features/Conversation/Conversations.cs
rpredescu-sinch: Hi @JPPortier!
I removed deprecated `Metadata` field.
Thank you!

tests/Sinch.Tests.Features/Conversation/Conversations.cs
rpredescu-sinch: Hi, @JPPortier!
This is an alias I have used because of namespace collision in the two test files.
It is a common pattern in dotnet, but I agree the code it is less confusing if we use the whole namespace instead (even if a slightly more verbose). As a consistent convention, we can always use full namespace in the case of this kind of conflicts. What do you think?
Thank you!

tests/Sinch.Tests/Conversation/ConversationsTests.cs
rpredescu-sinch: Hi, @JPPortier!
This is an alias I have used because of namespace collision in the two test files.
It is a common pattern in dotnet, but I agree the code it is less confusing if we use the whole namespace instead (even if a slightly more verbose). As a consistent convention, we can always use full namespace in the case of this kind of conflicts. What do you think?
Thank you!

tests/Sinch.Tests.Features/Conversation/Conversations.cs
JPPortier: We are in a test file, from my point of view it do not hurt to point not use alias anbd directly pointing to namespace+class

tests/Sinch.Tests.Features/Conversation/Conversations.cs
JPPortier: But the unused variable is still defined (and generate warnings): https://github.com/sinch/sinch-sdk-dotnet/blob/feature/DEVEXP-1255-conversation-conversations-e2e/src/Sinch/Conversation/Conversations/Conversation.cs#L21

rpredescu-sinch: Hi, @JPPortier.
Removed the unused member variable.
Thank you!

JPPortier (approved):
🚀 😍

PR #190, opened 2026-04-22
Support Conversation SinchEvents e2e

rpredescu-sinch:
## Summary
Support end-to-end tests for Conversation SinchEvents.

In this PR we:
- add e2e tests
- add unit tests


src/Sinch/Conversation/Webhooks/HmacAuthenticationValidation.cs
```diff
@@ -0,0 +1,114 @@
...
+    private static HMAC CreateHmac(string algorithm, string secret)
+    {
+        var secretBytes = Encoding.UTF8.GetBytes(secret);
+
+        return algorithm switch
+        {
+            "HmacSHA256" => new HMACSHA256(secretBytes),
+            _ => throw new CryptographicException($"Algorithm '{algorithm}' is not supported")
+        };
+    }
```
asein-sinch, line 113: At first I was a bit puzzled by this function and thought it would be better to reject the auth as soon as the header value was read. But considering the documentation "For now it's set to HmacSHA256.", it means the door is opened for other algorithms in the future. When it will be the case, this will be the single place to map the algorithm name and the crypto function 👍

src/Sinch/Conversation/Webhooks/Webhooks.cs
```diff
@@ -201,51 +205,16 @@ public Task Delete(string webhookId, CancellationToken cancellationToken = defau
...
-
-            using var hmac = new HMACSHA256(Encoding.UTF8.GetBytes(secret));
-            var hmacSha256 = hmac.ComputeHash(Encoding.UTF8.GetBytes(signedData));
-            var calculatedSignature = Convert.ToBase64String(hmacSha256);
-            _logger?.LogDebug("{CalculatedSignature}", calculatedSignature);
-            _logger?.LogDebug("{x-sinch-webhook-signature}", signature);
-            var isValidSignature = string.Equals(calculatedSignature, signature, StringComparison.Ordinal);
-            _logger?.LogInformation("The signature was validated with {success}", isValidSignature);
-            return isValidSignature;
+            return HmacAuthenticationValidation.ValidateAuthenticationHeader(secret, headers, body);
```
asein-sinch, line 211: Nice refactor offloading the function body 👍

tests/Sinch.Tests/Conversation/Webhooks/HmacAuthenticationValidationTests.cs
```diff
@@ -0,0 +1,167 @@
...
+            const string secret = "my_secret_key";
+            const string jsonPayload = "{\"event\":\"test\"}";
+            const string timestamp = "1736760161";
+            const string nonce = "01JHFFHWYY7HSS4FWTMDTQEK8V";
+            const string algorithm = "HmacSHA256";
+
+            const string toBeSigned = $"{jsonPayload}.{nonce}.{timestamp}";
+            using var hmac = new System.Security.Cryptography.HMACSHA256(Encoding.UTF8.GetBytes(secret));
+            var hash = hmac.ComputeHash(Encoding.UTF8.GetBytes(toBeSigned));
+            var signature = Convert.ToBase64String(hash);
```
asein-sinch: Just hardcode the expected value, don't reimplement the computing function (do the same wherever it applies)

tests/Sinch.Tests/Conversation/Webhooks/HmacAuthenticationValidationTests.cs
```diff
@@ -0,0 +1,167 @@
...
+                [NonceHeader] = ["n"],
+                [AlgorithmHeader] = ["HmacSHA256"],
+                [SignatureHeader] = ["s"]
+            };
+
+            headers.Remove(missingHeader);
+
+            var result = HmacAuthenticationValidation.ValidateAuthenticationHeader(secret, headers, jsonPayload);
+
+            result.Should().BeFalse();
```
asein-sinch, line 109: Here it can fail for many reasons, even if all the headers are present because you are using invalid values. I'd prefer you use the same valid that make the validation pass to really check the root cause are the missing headers.

tests/Sinch.Tests/Conversation/Webhooks/HmacAuthenticationValidationTests.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                [TimestampHeader] = [timestamp],
+                [NonceHeader] = [nonce],
+                [AlgorithmHeader] = [algorithm],
+                [SignatureHeader] = ["sig"]
+            };
+
+            var result = HmacAuthenticationValidation.ValidateAuthenticationHeader(secret, headers, jsonPayload);
+
+            result.Should().BeFalse();
```
asein-sinch, line 127: Same here, multiple causes of failure

tests/Sinch.Tests/Conversation/Webhooks/SinchEventsTests.cs
```diff
@@ -0,0 +1,161 @@
...
+    public class SinchEventsTests : ConversationTestBase
+    {
+        private const string CallbackSecret = "foo_secret1234";
+        private const string TimestampHeader = "x-sinch-webhook-signature-timestamp";
+        private const string NonceHeader = "x-sinch-webhook-signature-nonce";
+        private const string AlgorithmHeader = "x-sinch-webhook-signature-algorithm";
+        private const string SignatureHeader = "x-sinch-webhook-signature";
+
+        [Fact]
+        public void ValidateAuthenticationHeader_WithHeadersAndValidSignature_ReturnsTrue()
```
asein-sinch: I think you wanted to say "WithHeadersAndValidSecret"

tests/Sinch.Tests/Conversation/Webhooks/SinchEventsTests.cs
```diff
@@ -0,0 +1,161 @@
...
+                { SignatureHeader, ["wKmZBGo4Cf+y9cZoPHhiVw6ziKeubLGqN4OdG8jlaPo="] },
+            };
+
+            var result = Conversation.Webhooks.ValidateAuthenticationHeader(headers, payload, CallbackSecret);
+
+            result.Should().BeTrue();
+        }
+
+        [Fact]
+        public void ValidateAuthenticationHeader_WithHeadersAndInvalidSignature_ReturnsFalse()
```
asein-sinch: I think you wanted to say "WithHeadersAndInvalidSecret"

tests/Sinch.Tests/Conversation/Webhooks/SinchEventsTests.cs
asein-sinch, line 1: In this file, regarding the `ValidateAuthenticationHeader` method, I think we should focus on testing the different method signatures and outputs. The internal with algorithm check, secret valid/invalid, presence of headers, ... are already covered by the unit tests in HmacAuthenticationValidationTests

tests/Sinch.Tests.Features/Conversation/WebhooksEvents.cs
```diff
@@ -0,0 +1,248 @@
...
+        }
+    }
+
+    [Then(@"the Conversation event describes a FAILED event delivery status and its reason")]
+    public void ThenTheConversationEventDescribesAFailedEventDeliveryStatusAndItsReason()
+    {
+        var deliveryEvent = _parsedEvent.Should().BeOfType<DeliveryEvent>().Subject;
+        deliveryEvent.EventDeliveryReport.Should().NotBeNull();
+        deliveryEvent.EventDeliveryReport!.Status.Should().Be(DeliveryStatus.Failed);
+        deliveryEvent.EventDeliveryReport.Reason.Should().NotBeNull();
```
asein-sinch: Just for my knowledge, why do you need to assert not null `!` for `Status` and not for `Reason` ?

tests/Sinch.Tests.Features/Conversation/WebhooksEvents.cs
```diff
@@ -0,0 +1,248 @@
...
+            case "MESSAGE_SUBMIT":
+                {
+                    var messageSubmitEvent = _parsedEvent.Should().BeOfType<MessageSubmitEvent>().Subject;
+                    messageSubmitEvent.MessageSubmitNotification.Should().NotBeNull();
+                    messageSubmitEvent.MessageSubmitNotification!.SubmittedMessage.Should().NotBeNull();
+                    if (string.Equals(messageType, "media", StringComparison.OrdinalIgnoreCase))
+                    {
+                        messageSubmitEvent.MessageSubmitNotification.SubmittedMessage!.MediaMessage.Should().NotBeNull();
+                    }
+                    else
```
asein-sinch: Note: for these 2 `else`, I'd prefer to see a `if else` with the other message type explicitly mentioned, just to prepare the support for another one in the future.

tests/Sinch.Tests/Conversation/Webhooks/HmacAuthenticationValidationTests.cs
rpredescu-sinch: Hi @asein-sinch!
I have introduced a precomputed signature.
Thank you!

src/Sinch/Conversation/Webhooks/HmacAuthenticationValidation.cs
rpredescu-sinch: Hi @asein-sinch!
Indeed, the previous implementation ignored the `x-sinch-webhook-signature-algorithm` header entirely and always used `HMACSHA256`. Now we read and validate the header so if the Conversation API ever sends a different algorithm, we will throw a clear not implemented exception rather than introducing a wrong signature comparison.
Thank you!

src/Sinch/Conversation/Webhooks/Webhooks.cs
rpredescu-sinch: Thank you, @asein-sinch!
I did the same previously for SMS.
Best!

tests/Sinch.Tests.Features/Conversation/WebhooksEvents.cs
rpredescu-sinch: Hi @asein-sinch!
I have made the `else` branch explicit, thank you for pointing this out.
Best!

tests/Sinch.Tests.Features/Conversation/WebhooksEvents.cs
rpredescu-sinch: Good catch, @asein-sinch!
`!` in this context is the [C# null-forgiving operator](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving), and it suppresses a nullable warning (like "I know this isn't null, suppress the [CS8602 warning](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings#possible-dereference-of-null)"). However, it is not necessary here because the test project do not have this kind of warning enabled with `<Nullable>enable</Nullable>`, so no warnings will be produced anyway. I've removed the null-forgiving operator.
Thank you!

tests/Sinch.Tests/Conversation/Webhooks/HmacAuthenticationValidationTests.cs
rpredescu-sinch: Hi @asein-sinch!
True, thanks for catching this.
I have added a valid set of **valid variables** (including the precomputed signature) which when used together should guarantee a valid response. Now we have a success scenario covered, then by removing any of the required headers in a separate unit test, we know that it fails because of that specific header and nothing else. Same for the empty secret, all the other headers are valid together so the only reason for failure is the absence of the secret.
Best!

tests/Sinch.Tests/Conversation/Webhooks/HmacAuthenticationValidationTests.cs
rpredescu-sinch: See my comment here: https://github.com/sinch/sinch-sdk-dotnet/pull/190/changes#r3124244212
Thank you!

tests/Sinch.Tests/Conversation/Webhooks/SinchEventsTests.cs
rpredescu-sinch: Hi @asein-sinch!
I removed unnecessary tests we have already covered in `HmacAuthenticationValidationTests`.
Thank you!

tests/Sinch.Tests/Conversation/Webhooks/SinchEventsTests.cs
rpredescu-sinch: Hi, @asein-sinch!
Indeed! Now we have `ValidateAuthenticationHeader_EmptySecret_ReturnsFalse` in `HmacAuthenticationValidationTests.cs`
Thank you!

tests/Sinch.Tests/Conversation/Webhooks/SinchEventsTests.cs
rpredescu-sinch: Hi @asein-sinch!
Now we have it covered as `ValidateAuthenticationHeader_WithSingleValueHeaders_ReturnsTrue` test in `HmacAuthenticationValidationTests.cs`.
Thank you!

asein-sinch (approved):

PR #191, opened 2026-04-23
Conversation - Redesign of webhooks and callback configuration

rpredescu-sinch:
## Summary

Conversation "Event Destinations'" and "Sinch Events" terminology redesign.

## What it is in this PR

- Refactored `Hooks` to the new `SinchEvents` terminology
- `CallbackSettings` are now `EventDestinationSettings`


asein-sinch (commented):
Suggestion: the comment is generated from the spec, but since we are renaming the callback concepts, I think it will be more valuable to reference the SDK property instead `EventDestinationTarget` in the SendMessageRequest class. And in this class, as there is the mapping between the SDK property name and the API property name, the user will be able to relate both informations.
Let's keep it like this for the moment but note in the backlog that it may be worth to update all the SDK comments that are refering to the online doc, to instead refer to the our public SDK documentation when it will be online

src/Sinch/Conversation/Apps/Create/EventDestinationSettings.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <summary>
+    ///     This object contains additional settings related to event destination processing.
+    /// </summary>
+    public sealed class EventDestinationSettings
+    {
+        /// <summary>
+        ///     Secret can be used to sign contents of delivery receipts for a message that was sent with the default event destination target overridden (using the [&#x60;callback_url&#x60; field](https://developers.sinch.com/docs/conversation/api-reference/conversation/tag/Messages/#tag/Messages/operation/Messages_SendMessage!path&#x3D;callback_url&amp;t&#x3D;request)). You can then use the secret to verify the signature.
+        /// </summary>
+        [JsonPropertyName("secret_for_overridden_callback_urls")]
+        public string? SecretForOverriddenEventDestinationTarget { get; set; }
```
asein-sinch: Note: this variable is called `secretForOverriddenTarget` in the Java SDK. Double-check which one makes more sense as we are in a `EventDestinationSettings` object

src/Sinch/Conversation/Apps/Credentials/ConversationChannelCredentials.cs
```diff
@@ -99,12 +99,10 @@ public ConversationChannelCredentials(KakaoTalkChatCredentials kakaoTalkChatCred
 
 
         /// <summary>
-        ///     The secret used to verify the channel callbacks for channels which support callback verification. The callback verification is not needed for Sinch-managed channels because the callbacks are not leaving Sinch internal networks. Max length is 256 characters. Note: leaving channel_callback_secret empty for channels with callback verification will disable the verification.
+        ///     The secret used to verify the channel Sinch events for channels which support Sinch event verification. The Sinch event verification is not needed for Sinch-managed channels because the callbacks are not leaving Sinch internal networks. Max length is 256 characters. Note: leaving channel_callback_secret empty for channels with Sinch event verification will disable the verification.
```
asein-sinch: Note: there is still a reference to "callback". 
Note2: `channel_callback_secret` reference may not make sense in the SDK context, prefer referring to the class variable
Note3: the java SDK is still calling this variable "CallbackSecret" => @JPPortier

src/Sinch/Conversation/Apps/Credentials/ConversationChannelCredentials.cs
asein-sinch, line 1: If the "Credentials" namespace is used somewhere else than "Apps" it needs to be taken out of this folder

src/Sinch/Conversation/Apps/Update/UpdateAppRequest.cs
```diff
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.Text;
 using System.Text.Json.Serialization;
+using Sinch.Conversation.Apps.Create;
```
asein-sinch: The "update" should not depend on "create"

src/Sinch/Conversation/Apps/Update/UpdateAppRequest.cs
```diff
@@ -73,7 +80,7 @@ public sealed class UpdateAppRequest
         public override string ToString()
         {
             var sb = new StringBuilder();
-            sb.Append("class AppCreateRequest {\n");
+            sb.Append($"class {nameof(UpdateAppRequest)} {{\n");
```
asein-sinch, line 82: Good catch 👏

src/Sinch/Conversation/Apps/RateLimits.cs
```diff
@@ -23,11 +24,12 @@ public sealed class RateLimits
...
-        ///     The rate limit of callbacks sent to the webhooks registered for the app.
-        ///     Note that if you have multiple webhooks with shared triggers,
-        ///     multiple callbacks will be sent out for each triggering event. The default rate limit is 25.
+        ///     The rate limit of Sinch events sent to the event destinations registered for the app.
+        ///     Note that if you have multiple event destinations with shared triggers,
+        ///     multiple Sinch events will be sent out for each triggering event. The default rate limit is 25.
         /// </summary>
-        public long? Webhooks { get; set; }
+        [JsonPropertyName("webhooks")]
+        public long? EventDestinations { get; set; }
```
asein-sinch: I'm not sure we are talking about `EventDestinations` in this case but rather `Events`. This property must be renamed.
Note: as it is a long, does it make sense to suffix the property name with `Count` => @JPPortier to check with API standards
Note2: the Java SDK uses the property "Webhooks" => @JPPortier

src/Sinch/Conversation/Events/Send/SendEventRequest.cs
```diff
@@ -24,9 +25,10 @@ public sealed class SendEventRequest
 
 
         /// <summary>
-        ///     Overwrites the default callback url for delivery receipts for this message The REST URL should be of the form: &#x60;http://host[:port]/path&#x60;
+        ///     Overwrites the default event destination target for delivery receipts for this message The REST URL should be of the form: &#x60;http://host[:port]/path&#x60;
```
asein-sinch: Take the opportunity to reformat this comment. Also, as there is an explicit mention of using a URL, maybe it can be worth to reformulate with something like: "When the event destination target is a URL, it should be of the form..." (and also prefer a https protocol 🙈)

src/Sinch/Conversation/Messages/Send/SendMessageRequest.cs
```diff
@@ -30,13 +30,14 @@ public sealed class SendMessageRequest
 
 
         /// <summary>
-        ///     Overwrites the default callback url for delivery receipts for this message.
+        ///     Overwrites the default event destination target for delivery receipts for this message.
         ///     Note that you may
         ///     [define a `secret_for_overridden_callback_urls` at the app level](https://developers.sinch.com/docs/conversation/api-reference/conversation/tag/App/operation/App_UpdateApp!path=callback_settings/secret_for_overridden_callback_urls&amp;t=request)
         ///     this secret will be used to sign the contents of delivery receipts when the default
-        ///     callback URL is overridden by this property. The REST URL should be of the form: `http://host[:port]/path`
+        ///     event destination target URL is overridden by this property. The REST URL should be of the form: `http://host[:port]/path`
```
asein-sinch: Same comment as before

src/Sinch/Conversation/Messages/Send/SendMessageRequest.cs
```diff
@@ -30,13 +30,14 @@ public sealed class SendMessageRequest
...
         /// </summary>
-        public Uri? CallbackUrl { get; set; }
+        [JsonPropertyName("callback_url")]
+        public Uri? EventDestinationTarget { get; set; }
```
asein-sinch: Should it be a `Uri` or a `string`?

src/Sinch/Conversation/SinchEvents/Models/MessageInboundEventItem.cs
```diff
@@ -67,7 +67,7 @@ public sealed class MessageInboundEventItem
 
 
         /// <summary>
-        ///     Timestamp marking when the channel callback was received by the Conversation API.
+        ///     Timestamp marking when the channel Sinch event was received by the Conversation API.
```
asein-sinch: Here is really a "channel event" that the comment is referring to

src/Sinch/Conversation/SinchEvents/ConversationSinchEventBase.cs
```diff
@@ -12,7 +12,7 @@ public abstract class CallbackEventBase : ICallbackEvent
         public required string AppId { get; set; }
 
         /// <summary>
-        ///     Timestamp marking when the channel callback was accepted/received by the Conversation API.
+        ///     Timestamp marking when the channel Sinch event was accepted/received by the Conversation API.
```
asein-sinch: Here also, I think it should stay as "channel event".

tests/Sinch.Tests/Conversation/Apps/AppsTests.cs
```diff
@@ -677,7 +677,7 @@ private static ConversationChannelCredentials WithCallbackSecret(
             ConversationChannelCredentials credentials,
             string callbackSecret)
```
asein-sinch: You can also refactor the references to "callback" here too

MIGRATION_GUIDE.md
```diff
@@ -1049,8 +1054,8 @@ IEnumerable<Webhook> webhooks = await sinch.Conversation.Webhooks.List(appId);
 
 Version 2.*:
 ```csharp
-ListWebhooksResponse response = await sinch.Conversation.Webhooks.List(appId);
-IEnumerable<Webhook> webhooks = response.Webhooks ?? Enumerable.Empty<Webhook>();
+ListEventDestinationsResponse response = await sinch.Conversation.EventDestinations.List(appId);
+IEnumerable<EventDestination> destinations = response.EventDestinations ?? Enumerable.Empty<EventDestination>();
```
asein-sinch: You can name the variable `eventDestinations`. (Do the same for other variables named like this in this file)

MIGRATION_GUIDE.md
```diff
@@ -1149,5 +1154,142 @@ var callback = sinch.Conversation.Webhooks.ParseEvent(node!);
...
+Renamed types:
+
+- `ICallbackEvent` → `IConversationSinchEvent`
+- `CallbackEventBase` → `ConversationSinchEventBase`
+- `CallbackEventConverter` → `ConversationSinchEventConverter`
+- `UnsupportedCallbackEvent` → `UnsupportedConversationSinchEvent`
+
+Version 1.*:
+```csharp
+ICallbackEvent sinchEvent = sinch.Conversation.EventDestinations.ParseEvent(rawBody);
```
asein-sinch: `EventDestinations` didn't exist in v1, right? And even if the variable can be named the way we want, I don't think we advertized it as a `sinchEvent`

asein-sinch (changes requested):

src/Sinch/Conversation/Apps/Create/EventDestinationSettings.cs
rpredescu-sinch: Hi @asein-sinch!
I renamed it to `SecretForOverriddenTarget`, it makes more sense because as you said, we are already in a class named `EventDestinationSettings`, and it makes sense to avoid repetition of "EventDestination" in the property name.
Thank you!

src/Sinch/Conversation/Apps/Credentials/ConversationChannelCredentials.cs
rpredescu-sinch: Hi @asein-sinch!
`ConversationChannelCredentials` is only used inside apps so it make sense to keep it here.
Thank you!

src/Sinch/Conversation/Apps/Update/UpdateAppRequest.cs
rpredescu-sinch: Hi @asein-sinch!
Thanks for catching this!
I moved `EventDestinationSettings`, `DeliveryReportBasedFallback`, and `MessageRetrySettings` from `Sinch.Conversation.Apps.Create` to `Sinch.Conversation.Apps` because these types are shared by create, update, and read. Now `Update` namespace do not depend on the `Create` one.
Best!

src/Sinch/Conversation/Apps/Create/EventDestinationSettings.cs
JPPortier: Agree with @asein-sinch : we do not need to repeat "event destination" in this context

src/Sinch/Conversation/Events/Send/SendEventRequest.cs
rpredescu-sinch: Hi, @asein-sinch!

Enhanced `EventDestinationTarget` doc comments to:

```
///  Overwrites the default event destination target for delivery receipts for this event.
///  When the event destination target is a URL, it should be of the form: <c>https://host[:port]/path</c>
```

Also refactored the doc in same property of `SendMessageRequest.cs`.

src/Sinch/Conversation/Messages/Send/SendMessageRequest.cs
rpredescu-sinch: Hi @asein-sinch!
See the fix above:
https://github.com/sinch/sinch-sdk-dotnet/pull/191/changes#r3147343230
Thank you!

src/Sinch/Conversation/Messages/Send/SendMessageRequest.cs
rpredescu-sinch: Hi @asein-sinch !
Thanks for catching this!
The OAS defines `callback_url` as `string` in both `SendEventRequest` and `SendMessageRequest`. I'm changing it to `string` to align with the OAS and be consistent with `SendEventRequest`.
Best!

src/Sinch/Conversation/SinchEvents/Models/MessageInboundEventItem.cs
rpredescu-sinch: Hi, @asein-sinch !
Reverted the change.
Thank you!

tests/Sinch.Tests/Conversation/Apps/AppsTests.cs
rpredescu-sinch: Hi @asein-sinch !
Renamed test helpers from `callback` to `sinchEventSecret`:
- `WithCallbackSecret` to `WithSinchEventSecret`
- `WithCallbackSecretAndOrdinal` to `WithSinchEventSecretAndOrdinal`
- Parameter `callbackSecret` to `sinchEventSecret`

MIGRATION_GUIDE.md
rpredescu-sinch: Hi @asein-sinch!
Renamed variable from `destinations` to `eventDestinations`.
Thank you!

MIGRATION_GUIDE.md
rpredescu-sinch: Hi @asein-sinch!
That's true.
Restored `ICallbackEvent callbackEvent` for V1 migration.
Thank you!

src/Sinch/Conversation/SinchEvents/ConversationSinchEventBase.cs
rpredescu-sinch: Hi @asein-sinch !
Reverted xml documentation to `channel event`.
Best!

src/Sinch/Conversation/Apps/Credentials/ConversationChannelCredentials.cs
JPPortier: Not sure we are talking about Sinch Event/Destination rights ?

Here
- it is at Channel Credential level (and specs says: `The secret used to verify the channel callbacks for channels which support callback verification. The callback verification is not needed for Sinch-managed channels because the callbacks are not leaving Sinch internal networks.`)
- not https://developers.sinch.com/docs/conversation/api-reference/conversation/app/app_updateapp#app/app_updateapp/t=request&path=callback_settings
- nor https://developers.sinch.com/docs/conversation/api-reference/conversation/webhooks/webhooks_createwebhook#webhooks/webhooks_createwebhook/t=request&path=secret

src/Sinch/Conversation/Apps/Update/UpdateAppRequest.cs
JPPortier: Right.

`EventDestinationSettings is no longer used from 'Create' namespace only

src/Sinch/Conversation/Apps/RateLimits.cs
JPPortier: 1. Right, more "SinchEvents" related
2. This is a V1 of conversation so not tied to standards but at SDK level we can improve access to schema yes. Then also having to rename the 3 fields of RateLimits to have `Count` suffix used.
3. Shame on Java SDK !

src/Sinch/Conversation/EventDestinations/EventDestinations.cs
JPPortier, line 1: - Here path directory is `EventDestinations`. Numbers related PR is using `EventDestination` not in plural form.
- Namespace here is also plural form but not in Numbers.
- Here interface is named ISinchConversationEventDestinations but ISinchNumbersEventDestination onto numbers (not plural form)

Could you check, and unified them if having too, for other opened PRs please ?

src/Sinch/Conversation/EventDestinations/ListEventDestinationsResponse.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Sinch.Conversation.EventDestinations
+{
+    public sealed class ListEventDestinationsResponse
+    {
+        /// <summary>
+        ///     List of event destinations belonging to a specific project ID and app ID.
+        /// </summary>
+        [JsonPropertyName("webhooks")]
+        public IEnumerable<EventDestination>? EventDestinations { get; init; }
```
JPPortier: Should be defined as `List`

src/Sinch/Conversation/EventDestinations/EventDestinations.cs
```diff
@@ -0,0 +1,248 @@
...
+                    cancellationToken);
+            if (jsonResult == null)
+            {
+                _logger?.LogError("Failed to deserialize conversation Sinch event. No matching event type found.");
+                throw new InvalidOperationException("Deserialization of conversation Sinch event failed");
+            }
+
+            return jsonResult;
+        }
+    }
```
JPPortier, line 247: This should be related to Event Destination configuration but at "SinchEvents" subdomains level

Then end user will be able to use SinchEvents without having to configure Conversation with unused but required parameters.

➡️ Missing part with this PR/ticket: https://github.com/sinch/sinch-sdk-dotnet/blob/feature/DEVEXP-1311-conversation-event-destinations-and-sinch-events/src/Sinch/SinchClient.cs#L278

src/Sinch/Conversation/SinchEvents/ConversationSinchEventConverter.cs
```diff
@@ -180,12 +180,12 @@ public override void Write(Utf8JsonWriter writer, ICallbackEvent value, JsonSeri
                 case OptOutEvent optOutEvent:
                     JsonSerializer.Serialize(writer, optOutEvent, options);
                     break;
-                case UnsupportedCallbackEvent unsupportedEvent:
+                case UnsupportedConversationSinchEvent unsupportedEvent:
```
JPPortier: Regarding https://developers.sinch.com/docs/conversation/callbacks#unsupported-callback keeping `UnsupportedCallbackEvent` still have sense or we have to rename all SinchEvents family classes with `ConversationSinchEvent` suffix (OptOutConversationSinchEvent, DeleteConversationSinchEvent, ...)

tests/Sinch.Tests.Features/Conversation/SinchEvents.cs
```diff
@@ -3,29 +3,29 @@
...
     private string _rawEvent;
-    private ICallbackEvent _parsedEvent;
+    private IConversationSinchEvent _parsedEvent;
 
     [Given(@"the Conversation Webhooks handler is available")]
-    public void GivenTheConversationWebhooksHandlerIsAvailable()
+    public void GivenTheConversationEventDestinationsHandlerIsAvailable()
     {
-        _webhooks = Utils.SinchConversationClient().Webhooks;
+        _eventDestinations = Utils.SinchConversationClient().EventDestinations;
```
JPPortier, line 28: This should be a `Utils.SinchConversationClient().SinchEvents` and using a SinchEvents inited without any required parameters to init Conversation domains

src/Sinch/Conversation/EventDestinations/EventDestinations.cs
rpredescu-sinch: Updated in Numbers to plural form.

src/Sinch/Conversation/EventDestinations/ListEventDestinationsResponse.cs
rpredescu-sinch: Updated it to List. We also have a separate ticket to update request/response models `IEnumerable` properties to `List`.

src/Sinch/Conversation/Apps/RateLimits.cs
rpredescu-sinch: Hi @asein-sinch!
Renamed it to `EventsCount`, thank you for the suggestion.
Best!

src/Sinch/Conversation/SinchEvents/ConversationSinchEventConverter.cs
rpredescu-sinch: Reverted `UnsupportedConversationSinchEvent` to original `UnsupportedCallbackEvent` for consistency with the other event models.

tests/Sinch.Tests.Features/Conversation/SinchEvents.cs
rpredescu-sinch: Hi @JPPortier.

See my comment and fix above: https://github.com/sinch/sinch-sdk-dotnet/pull/191#discussion_r3152579733

src/Sinch/Conversation/EventDestinations/EventDestinations.cs
rpredescu-sinch: Hi @JPPortier.

This is a refactoring I'll apply in a separate PR. Will update here once ready.

**Update:** `ConversationSinchEvents` can now be instantiated directly, without going through `SinchClient` at all so users who only need `ParseEvent` or `ValidateAuthenticationHeader` don't need to configure a region or credentials. See the changes in this dedicated PR: https://github.com/sinch/sinch-sdk-dotnet/pull/198

src/Sinch/Conversation/Apps/Credentials/ConversationChannelCredentials.cs
rpredescu-sinch: Hi @asein-sinch!
Thank for catching this.
I updated XML documentation: `channel_callback_secret` is now `<see cref="SinchEventSecret"/>`  and replaced `callbacks` to `events`. 
Best!

rpredescu-sinch:
> Suggestion: the comment is generated from the spec, but since we are renaming the callback concepts, I think it will be more valuable to reference the SDK property instead `EventDestinationTarget` in the SendMessageRequest class. And in this class, as there is the mapping between the SDK property name and the API property name, the user will be able to relate both informations. Let's keep it like this for the moment but note in the backlog that it may be worth to update all the SDK comments that are refering to the online doc, to instead refer to the our public SDK documentation when it will be online

@asein-sinch 
I have created a ticket to handle this separately, thank you!

PR #194, opened 2026-04-24
Verification - Redesign of webhooks and callback configuration

rpredescu-sinch:
## Summary

Verification "Event Destinations'" and "Sinch Events" terminology redesign.

## What it is in this PR

- Refactored `Hooks` to the new `SinchEvents` terminology
- Updated callback to Sinch event

The refactor affects folders, namespaces, source files, class properties, and tests.


src/Sinch/Verification/SinchEvents/FlashCallRequestSinchEventResponse.cs
```diff
@@ -1,8 +1,8 @@
 using System.Text.Json.Serialization;
 
-namespace Sinch.Verification.Hooks
+namespace Sinch.Verification.SinchEvents
 {
-    public sealed class FlashCallRequestEventResponse : RequestEventResponseBase
+    public sealed class FlashCallRequestSinchEventResponse : RequestSinchEventResponseBase
```
asein-sinch: The classname is quite difficult to understand: what about `VerificationStartEventResponseFlashCall` to indicate this is a "response" to a "verification start" action with "flash call" as a medium?
 - FlashCallRequestSinchEventResponse -> VerificationStartEventResponseFlashCall
 - PhoneCallRequestSinchEventResponse -> VerificationStartEventResponsePhoneCall
 - SmsRequestSinchEventResponse -> VerificationStartEventResponseSms
 - WhatsAppRequestSinchEventResponse -> VerificationStartEventResponseWhatsApp
On top of improving the meaning, it will bring the advantage of grouping all the classes next to each other in the folder.

src/Sinch/Verification/SinchEvents/RequestSinchEventResponseBase.cs
```diff
@@ -1,8 +1,8 @@
 using System.Text.Json.Serialization;
 
-namespace Sinch.Verification.Hooks
+namespace Sinch.Verification.SinchEvents
 {
-    public class RequestEventResponseBase
+    public class RequestSinchEventResponseBase
```
asein-sinch: This one can be renamed too 😉

src/Sinch/Verification/SinchEvents/VerificationSinchEventRequest.cs
```diff
@@ -2,14 +2,14 @@
...
 {
     /// <summary>
-    ///     This callback event is a POST request to the specified verification callback URL and
+    ///     This Sinch event is a POST request to the specified verification event destination URL and
     ///     is triggered when a new verification request is made from the SDK client or the Verification Request API.
-    ///     This callback event is only triggered when a verification callback URL is specified in your dashboard.
+    ///     This Sinch event is only triggered when a verification event destination URL is specified in your dashboard.
     /// </summary>
-    public sealed class VerificationRequestEvent
+    public sealed class VerificationSinchEventRequest
```
asein-sinch: For this class, I'd prefer to see `VerificationStartEvent`. It's true that technically, this is a POST request that will land on your callback server, but this is the case for all kinds of events, there is no need to mention the word "Request" in the classname (this is coming from the OAS schema name).
What matters is that when a callback URL is specified, this event will be fired when a verification process is started.

src/Sinch/Verification/SinchEvents/VerificationSinchEventResult.cs
```diff
@@ -1,15 +1,15 @@
...
     /// <summary>
-    ///     This callback event is a POST request to the specified verification callback URL and triggered when
+    ///     This Sinch event is a POST request to the specified verification event destination URL and triggered when
     ///     a verification has been completed and the result is known.
     ///     It's used to report the verification result to the developer's backend application.
-    ///     This callback event is only triggered when the verification callback URL is specified in your dashboard.
+    ///     This Sinch event is only triggered when the verification event destination URL is specified in your dashboard.
     /// </summary>
-    public sealed class VerificationResultEvent
+    public sealed class VerificationSinchEventResult
```
asein-sinch: And for this one `VerificationResultEvent` was fine ☺️

src/Sinch/Verification/SinchEvents/FlashCallRequestSinchEventResponse.cs
rpredescu-sinch: Nice suggestion, @asein-sinch!
Definitely much better than having "Response" and "Request" in the same name, with "Sinch" in between.
Thank you!

src/Sinch/Verification/SinchEvents/RequestSinchEventResponseBase.cs
rpredescu-sinch: Hi @asein-sinch!
Followed the same naming pattern and refactored it to `VerificationStartEventResponseBase`.
Thank you!

src/Sinch/Verification/SinchEvents/VerificationSinchEventRequest.cs
rpredescu-sinch: Hi @asein-sinch!
`VerificationStartEvent` makes more sense indeed, thank you for the suggestion, updated the class name.
Best!

src/Sinch/Verification/SinchEvents/VerificationSinchEventResult.cs
rpredescu-sinch: @asein-sinch 
Oh, reverted it to the original name and removed it from migration guide as it stays the same.
Thank you!

asein-sinch (approved):

PR #200, opened 2026-05-06
Add explicit solution file builds to catch stale project paths

rpredescu-sinch:
## Summary

The `build.yaml` workflow uses dotnet build `examples/dirs.proj` (a traversal project) to build all example `.csproj` files. However, this only validates that individual projects compile, but it does not validate that the `.sln` files (`Snippets.sln`, `Webhook.Template.sln`) correctly reference those projects.

When folders are renamed (e.g., callbacks/ → event-destinations/), the `.csproj` files move but the `.sln` retains stale paths. The traversal project still finds and builds the` .csproj` at its new location, so CI passes, but the solution file is broken for anyone opening it in an IDE.

## Root Cause
`dirs.proj` discovers `.csproj` files by glob (`**\*.csproj`), completely bypassing solution-level path references.

## Fix
Add two new CI steps to cover solution file builds after the existing Build Examples step.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -95,6 +95,12 @@ jobs:
       - name: Build Examples
         run: dotnet build examples/dirs.proj
 
+      - name: Build Snippets Solution
+        run: dotnet build examples/snippets/Snippets.sln
+
+      - name: Build Webhook Template Solution
+        run: dotnet build examples/templates/webhooks/Webhook.Template.sln
+
       - name: Test
         run: dotnet test tests/Sinch.Tests --framework ${{ matrix.tfm }} --no-restore --verbosity normal
 


.github/workflows/build.yaml
```diff
@@ -95,6 +95,12 @@ jobs:
       - name: Build Examples
         run: dotnet build examples/dirs.proj
 
+      - name: Build Snippets Solution
```
JPPortier, line 98: Is a "master" .sln file could contains references multiples "children" .sln files ?

If yes, idea here is to have a common file for both CI validation but also offer to any .sln file user view of all examples content in one file ➡️ DX

rpredescu-sinch: A `.sln` file can’t directly contain or reference other `.sln` files unfortunately, only projects. We could add all the projects in templates and snippets in one solution file and organize them in separate solution folders (while still having one solution file), but I think what we have now is nicer and a good isolation.

JPPortier: Ok.
We can stay with different .sln files.
Just in the future, if/when we add another .sln file (.e.g. getting stated) we jabe to not forget to update CI

JPPortier (approved):

PR #199, opened 2026-05-04
Numbers - Redesign of webhooks and callback configuration

rpredescu-sinch:
## Summary

Numbers to "Event Destinations'" and "Sinch Events" terminology redesign.

## What it is in this PR

- Refactored `Webhooks` to the new `EventDestinations` terminology
- `Sinch.Numbers.Hooks`  namespace renamed to `Sinch.Numbers.SinchEvents`
- `CallbackConfiguration` are now `EventDestination`
- `CallbackUrl` is no `EventDestinationTarget`
- SinchClient lazily check the credentials when executing a Numbers method
- New template showcasing the handling of an event for the Numbers Events

The refactor affects folders, namespaces, source files and class properties, tests.

PR #201, opened 2026-05-07
SMS - Redesign of webhooks and callback configuration

rpredescu-sinch:
## Summary

SMS to "Event Destinations'" and "Sinch Events" terminology redesign.

## What it is in this PR

- `Sinch.SMS.Hooks`  namespace renamed to `Sinch.SMS.SinchEvents`
- `CallbackUrl` is no `EventDestinationTarget`
- SinchClient lazily check the credentials when executing a SMS method


examples/templates/sinch-events/Sms/SmsSinchEventsController.cs
```diff
@@ -1,25 +1,24 @@
...
+using Sinch.SMS.SinchEvents;
 using SinchEvents.Template;
 
 namespace SinchEvents.Template.Sms;
 
 /// <summary>
 /// Example controller for receiving SMS Sinch Events.
 /// Set requireAuthentication to true once you have configured a shared secret
-/// at 'Sinch:Sms:WebhookSecret' in appsettings.
-/// See https://developers.sinch.com/docs/sms/api-reference/sms/tag/Webhooks/
```
JPPortier, line 11: Direct link to online doc is still having sense for end users

examples/templates/sinch-events/appsettings.json
```diff
@@ -6,11 +6,8 @@
     }
   },
   "Sinch": {
-    "KeyId": "",
-    "KeySecret": "",
-    "ProjectId": "",
     "Sms": {
-      "WebhookSecret": ""
+      "SinchEventSecret": ""
```
JPPortier: Better to stay consistent with plural form (expect when talking about individual event which is not the case here)
```suggestion
      "SinchEventsSecret": ""
```

examples/templates/sinch-events/appsettings.json
```diff
@@ -6,11 +6,8 @@
...
     },
     "Numbers": {
       "HmacSecret": ""
```
JPPortier: Better to align naming to have a consistent ones across domains: `SinchEventsSecret` for all of them seems to be a good choice

examples/templates/sinch-events/Program.cs
```diff
@@ -14,19 +14,10 @@
...
-builder.Services.AddSinchClient(() => new SinchClientConfiguration
-{
-    SinchUnifiedCredentials = new SinchUnifiedCredentials
-    {
-        ProjectId = builder.Configuration["Sinch:ProjectId"]!,
-        KeyId = builder.Configuration["Sinch:KeyId"]!,
-        KeySecret = builder.Configuration["Sinch:KeySecret"]!
-    }
-});
+builder.Services.AddSinchClient(() => new SinchClientConfiguration());
```
JPPortier: SinchClient is now having a constructor without any parameters for configuration

Is the empty `new SinchClientConfiguration()` is required here ?

src/Sinch/SMS/Batches/Send/BatchBase.cs
```diff
@@ -104,16 +104,16 @@ public abstract class BatchBase
...
-        ///     Override the default callback URL for this batch. Must be a valid URL.
-        ///     Learn how to set a default callback URL
+        ///     Override the default event destination URL for this batch. Must be a valid URL.
+        ///     Learn how to set a default event destination target URL
         ///     <see href="https://community.sinch.com/t5/SMS/How-do-I-assign-a-callback-URL-to-an-SMS-service-plan/ta-p/8414">here</see>
-        ///     .
         /// </summary>
-        public Uri? CallbackUrl { get; set; }
+        [JsonPropertyName("callback_url")]
+        public Uri? EventDestinationTarget { get; set; }
```
JPPortier: Within previous PR, it was validated to use `string` but not `Uri`: https://github.com/sinch/sinch-sdk-dotnet/pull/191#discussion_r3143297032

JPPortier:
### Code review

Found 1 issue:

1. Missing test for direct `SmsSinchEvents` instantiation without `SinchClient` — the requirements for DEVEXP-1311 explicitly state: *"new Numbers.SinchEvents().parseEvent(body) / new Sms.SinchEvents().parseEvent(body) — This other way of calling the SinchEvent methods should be tested in an integration test"*. PR #199 (Numbers) includes a `ParseEvent_WorksWithDirectInstantiation` test that directly instantiates `new NumbersSinchEvents(options)` without going through `SinchClient`. No equivalent test exists in the SMS test file.

https://github.com/sinch/sinch-sdk-dotnet/blob/fce99c1deec6f831d884a29c12573872a31ec95d/tests/Sinch.Tests/Sms/SinchEventsTests.cs#L1-L30

Numbers counterpart for reference (PR #199, now merged): https://github.com/sinch/sinch-sdk-dotnet/blob/fce99c1deec6f831d884a29c12573872a31ec95d/tests/Sinch.Tests/Numbers/SinchEventsTests.cs#L28-L38

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/Sinch/SMS/SinchEvents/SmsSinchEvents.cs
```diff
@@ -0,0 +1,74 @@
...
+            ILoggerAdapter<ISmsSinchEvents>? logger = null)
+        {
+            _jsonSerializerOptions = jsonSerializerOptions;
+            _logger = logger;
+        }
+
+        /// <inheritdoc />
+        public ISmsSinchEvent ParseEvent(string json)
+        {
+            var result = JsonSerializer.Deserialize<ISmsSinchEvent>(json, _jsonSerializerOptions) as ISmsSinchEvent;
```
JPPortier: From Numbers counterpart, function is:
```
var result = JsonSerializer.Deserialize<NumbersSinchEvent>(json, _jsonSerializerOptions);
```
Is pattern have to be different ?

src/Sinch/SMS/SinchEvents/HeaderValidation.cs
JPPortier, line 1: Within Numbers, this feature is within SinchEvents directory and file is named "HeaderValidation"

What do think about following a common pattern ?

src/Sinch/SMS/SmsClient.cs
```diff
@@ -71,13 +61,32 @@ internal record ProjectId(string Value);
 
     internal sealed class SmsClient : ISinchSms
     {
+        private readonly ISinchSmsBatches? _batches;
+        private readonly ISinchSmsDeliveryReports? _deliveryReports;
+        private readonly ISinchSmsGroups? _groups;
+        private readonly ISinchSmsInbounds? _inbounds;
+        private readonly bool _supportsApiOperations;
```
JPPortier: Regarding now lazy loading is in place and requests should be validated at request level at least for unified credentials, is this variable is still required and/or having same behavior in service plan id context ?

src/Sinch/SinchClient.cs
```diff
@@ -359,23 +359,24 @@ private SmsClient InitSms()
...
             }
 
-            var unifiedCredentials = ValidateUnifiedCredentials();
-
             if (sinchSmsConfiguration.Region == null)
-                throw new InvalidOperationException(
-                    $"{nameof(SinchSmsConfiguration)}.{nameof(SinchSmsConfiguration.Region)} is required. " +
-                    $"Set it to one of the values in {nameof(SmsRegion)}, e.g. {nameof(SmsRegion)}.{nameof(SmsRegion.Us)}.");
+            {
+                _logger?.LogInformation("Initializing SMS Sinch Events without outbound SMS region configuration");
```
JPPortier: This kind of warning is not required if in SinchEvents init: this an expected use case.
But: not seeing how having region to null enable to detect we are in SinchEvents init only phase ?
When the validation onto region as required will now be triggered at runtime before calling any SMS rest API endpoint ?

tests/Sinch.Tests.Features/Sms/SinchEvents.cs
```diff
@@ -30,61 +28,61 @@ public class Webhooks
         private string _rawRecipientDeliveryReportAbortedContent;
 
         [Given(@"the SMS Webhooks handler is available")]
-        public void GivenTheSmsWebhooksHandlerIsAvailable()
+        public void GivenTheSmsSinchEventsHandlerIsAvailable()
         {
-            _webhooks = Utils.SinchClient.Sms.Webhooks;
+            _smsSinchEvents = Utils.SinchClient.Sms.SinchEvents;
```
JPPortier: e2e test for SinchEvents must use a SinchCLient inited without configuration

tests/Sinch.Tests.Features/Sms/SinchEvents.cs
```diff
@@ -1,25 +1,23 @@
...
 
 namespace Sinch.Tests.Features.Sms
 {
     [Binding]
-    public class Webhooks
+    public class SinchEvents
     {
-        private const string WebhookSecret = "KayakingTheSwell";
-        private const string WebhooksUrlPrefix = "http://localhost:3017/webhooks/sms";
+        private const string SinchEventSecret = "KayakingTheSwell";
```
JPPortier:
```suggestion
        private const string SinchEventsSecret = "KayakingTheSwell";
```

MIGRATION_GUIDE.md
```diff
@@ -511,7 +507,9 @@ The following source files were removed as part of the refactor and should be de
 
 Replacement guidance
 
-- If you previously relied on `IIncomingSms` or `IncomingBinarySms`, switch to the new strongly-typed webhook models (`TextMessage`, `BinaryMessage`, etc.) and prefer the `ISmsWebhooks.ParseEvent(string json)` helper which returns an `ISmsEvent` that you can pattern-match or cast as shown above.
+- If you previously relied on `IIncomingSms`, switch to `IInbound`.
+- If you previously relied on `IncomingTextSms` or `IncomingBinarySms`, switch to `SmsInbound` or `BinaryInbound`.
+- Prefer the `ISmsSinchEvents.ParseEvent(string json)` helper, which now returns `ISmsSinchEvent` that you can pattern-match or cast to `IInbound`, `BatchDeliveryReportSms`, `RecipientDeliveryReportSms`, and the other concrete Sinch Events models.
```
JPPortier: > Prefer 

There is no other way right ? deserilization option are not public, so end user **must** use parse event to have proper output. no ?

MIGRATION_GUIDE.md
```diff
@@ -522,32 +520,104 @@ var bin = JsonSerializer.Deserialize<IncomingBinarySms>(json);
...
 ```csharp
-var bin = sinchClient.Sms.Webhooks.ParseEvent(json).As<BinaryMessage>();
+var bin = sinchClient.Sms.SinchEvents.ParseEvent(json).As<BinaryInbound>();
 // or when using plain deserialization:
-var bin = JsonSerializer.Deserialize<BinaryMessage>(json);
+var bin = JsonSerializer.Deserialize<BinaryInbound>(json);
 ```
 
-## SMS Webhooks: property rename for recipient delivery reports
+## SMS Webhooks: property rename for per-recipient delivery reports
```
JPPortier: > SMS Webhooks

`webhooks` ?

examples/templates/sinch-events/Sms/SmsSinchEventsController.cs
rpredescu-sinch: Added back the documentation link.

examples/templates/sinch-events/appsettings.json
rpredescu-sinch: Using plural form `SinchEventsSecret`.

examples/templates/sinch-events/appsettings.json
rpredescu-sinch: Using `SinchEventsSecret` for Numbers and will use the same for other domains.

examples/templates/sinch-events/Program.cs
rpredescu-sinch: This is the ASP.NET dependency injection registration pattern, not direct `SinchClient` instantiation. Because we already have a parameter-less `SinchClient`, I have introduced an `AddSinchClient()` overload which does not requires configuration details, and can be used to parse and validate Sinch Events.

src/Sinch/SMS/Batches/Send/BatchBase.cs
rpredescu-sinch: Moved the property type from `Uri` to `string` and will keep this in mind for the other domains.

src/Sinch/SMS/SinchEvents/SmsSinchEvents.cs
rpredescu-sinch: I removed  unnecessary `ISmsSinchEvent` cast if you were referring to it.

Numbers deserializes to a concrete type in `NumbersSinchEvent`, while SMS deserializes to the interface because the converter is attached directly to `ISmsSinchEvent`. `System.Text.Json` does not support deserializing interface or abstract classes unless we provide a converter. SMS interface has one, Numbers currently does not.

src/Sinch/SMS/SinchEvents/HeaderValidation.cs
rpredescu-sinch: Moved `HmacAuthenticationValidation` to `SinchEvents` and renamed it `HeaderValidation`.

tests/Sinch.Tests.Features/Sms/SinchEvents.cs
rpredescu-sinch: Renamed it to plural form.

src/Sinch/SMS/SmsClient.cs
rpredescu-sinch: Removed the previous “supports API operations” flag and validation happens lazily at subdomain level. 

If `ServicePlanIdRegion` is missing, `SinchEvents` still work, but API access fails with an exception. We can extend validation in the future and add more checks. 

The separation between **API** and **SinchEvents** scenarios is still not very obvious in the code and further refactoring might be needed.

Now the behavior is the correct one and validation returns the right message when trying to consume the Sinch API without proper configuration or region.

src/Sinch/SinchClient.cs
rpredescu-sinch: You are right, `Region == null` is not a good check for a missing configuration. It means more that we do not have enough configuration to create the SMS subdomains during initialization, so we use the `SmsClient` constructor which allows the consumer to use `SinchEvents` at least.

The actual validation is now deferred and happens when one of the subdomains are accessed (`Batches, Inbounds, ...). and a region validation error is now triggered at runtime when using subdomains without proper initialization.

tests/Sinch.Tests.Features/Sms/SinchEvents.cs
rpredescu-sinch: Using `new SinchClient().Sms.SinchEvents;`.

MIGRATION_GUIDE.md
rpredescu-sinch: That is correct. 

`ISmsSinchEvents.ParseEvent(string json)` is the only public API that resolves an SMS Sinch Events payloads to the concrete `ISmsSinchEvent` implementation.

I updated the documentation to:

"Prefer `ISmsSinchEvents.ParseEvent(string json)` when the payload type is not known in advance. It is the public API that resolves the incoming SMS Sinch Event JSON to the correct concrete `ISmsSinchEvent` implementation. If you already know the concrete payload type, you can still deserialize directly to that concrete model."

MIGRATION_GUIDE.md
rpredescu-sinch: Added "SinchEvent" alongside "Webhooks" terminology.

src/Sinch/SMS/SinchEvents/SmsSinchEvents.cs
JPPortier: > I removed unnecessary ISmsSinchEvent cast if you were referring to it.

Rgith.
There was here an `as  ISmsSinchEvent` which was ntore required within Numbers

src/Sinch/SMS/Batches/Send/BatchBase.cs
JPPortier: There is still a reference to ‘Uri’ in the SMS related code: https://github.com/sinch/sinch-sdk-dotnet/blob/feature/DEVEXP-1311-sms-event-destinations-and-sinch-events-v2/src/Sinch/SMS/Batches/Send/SendMediaBatchRequest.cs#L66

src/Sinch/SMS/SmsClient.cs
```diff
@@ -71,13 +61,35 @@ internal record ProjectId(string Value);
...
     {
+        private readonly ISinchSmsBatches? _batches;
+        private readonly ISinchSmsDeliveryReports? _deliveryReports;
+        private readonly ISinchSmsGroups? _groups;
+        private readonly ISinchSmsInbounds? _inbounds;
+        private readonly ServicePlanIdConfiguration? _servicePlanIdConfiguration;
+
+        /// <summary>
+        ///     Creates an SMS client that supports Sinch Events parsing and signature validation only.
+        ///     SMS API operations remain unavailable until SMS configuration is provided.
```
JPPortier: > SMS API operations remain unavailable until SMS configuration is provided.

But this constructor is accepting a `ServicePlanIdConfiguration`? Not used if only for SinchEvents purpose right ?

src/Sinch/SMS/SmsClient.cs
JPPortier: > The separation between API and SinchEvents scenarios is still not very obvious in the code and further refactoring might be needed.

What do you mean by `further refactoring might be needed` ?
What is missing with the current PR ?

src/Sinch/SinchClient.cs
JPPortier: There is still SMS related validation at SinchClient level.
Only SMS should perform the logic about what is required and what is not.
During discussion last wee agreed to have SMS validate inputs when really needed.
Here, any SDK user could have log onto his system about region even if use case do not required region settings (SinchEvents usage only).

Let SMSClient check at it own level what is required and what is not. In SinchEvents context: nothing is required and do not need send unrelated logs

src/Sinch/Conversation/Webhooks/HeaderValidation.cs
```diff
@@ -9,7 +9,7 @@ namespace Sinch.Conversation.Webhooks;
 /// <summary>
 ///     Validates HMAC authentication headers for Conversation webhook requests.
 /// </summary>
-internal sealed class HmacAuthenticationValidation
+internal sealed class HeaderValidation
```
JPPortier, line 12: Conversation was not in the scope of this PR, but OK, you're ahead of time :-)

src/Sinch/SMS/SmsClient.cs
```diff
@@ -45,22 +45,12 @@ public interface ISinchSms
...
-        /// </summary>
-        /// <remarks>
-        ///     SMS webhooks are configured in the Sinch Dashboard, not via API.
-        ///     This service only handles parsing and validation of incoming webhook requests.
-        /// </remarks>
-        ISmsWebhooks Webhooks { get; }
+        /// <inheritdoc cref="ISmsSinchEvents" />
+        ISmsSinchEvents SinchEvents { get; }
 
         internal bool IsUsingServicePlanId { get; }
```
JPPortier: It is no longer used right ?

src/Sinch/SMS/SmsClient.cs
```diff
@@ -102,35 +114,65 @@ internal SmsClient(ServicePlanId servicePlanId, Uri baseAddress, LoggerFactory?
...
+
+        private T GetRequiredApiClient<T>(T? client)
+            where T : class
+        {
+            if (client != null)
+                return client;
+
+            ValidateApiClientAvailability();
+
+            throw CreateApiConfigurationException();
```
JPPortier: Curious about .NET 
Here if `ValidateApiClientAvailability` do not throw any exception we will throw a InvalidOperationException even if no credentials are passed.

How does it work ?

src/Sinch/Conversation/Webhooks/HeaderValidation.cs
rpredescu-sinch: Got excited by refactoring SMS and touched Conversation by accident. :)

src/Sinch/SMS/SmsClient.cs
rpredescu-sinch: I was only thinking if the four constructor overloads which made `SmsClient` initialization difficult to understand, but I was shy to introduce this refactoring in the scope of this PR.

In the end, I went and consolidated constructors after you questioned remaining SMS business logic in `SinchClient` in a comment below.

src/Sinch/SinchClient.cs
rpredescu-sinch: Thanks for pushing on this.

Addressing it required consolidating the four `SmsClient` constructor overloads, which I had initially left out of scope for this PR. 

In the end I went for it, and the result is cleaner than expected: a single constructor, no SMS logic in `SinchClient`, no unrelated logs for SinchEvents only users.

src/Sinch/SMS/SmsClient.cs
rpredescu-sinch: The multiple constructor overloads made it difficult to properly initialize the right `SmsClient` in `SinchClient`. 
The parameter wasn't used for initialization, but to choose which error to throw when an API property was accessed. 

That was a sign the constructors design was flawed: it was mixing construction concerns with error reporting concerns, and that's why I wrote in a comment above that a future refactoring was needed.

However, after the last refactoring that I have introduced, we have one constructor left and things are much simpler.

src/Sinch/SMS/SmsClient.cs
rpredescu-sinch: It was set in the constructor but never read anywhere. Dead code, I removed it.

src/Sinch/SMS/SmsClient.cs
rpredescu-sinch: The compiler required the second throw to be there because `ValidateApiClientAvailability` returns `void` and it doesn't know if the method throw an exception or not. From the compiler's perspective, execution could always continue past it, so without the second throw the method would have a code path that returns nothing for a non-void method, causing a compile error.

This is why the design was awkward: two separate throw paths in sequence, where only one could actually fire. The refactoring we did now eliminates this situation for good.

src/Sinch/SMS/SmsClient.cs
```diff
@@ -45,92 +47,83 @@ public interface ISinchSms
...
+        {
+            return client ?? throw CreateApiConfigurationException();
+        }
+
+        private static InvalidOperationException CreateApiConfigurationException()
+        {
+            return new InvalidOperationException(
+                "SMS API operations are unavailable when the SMS client is initialized for Sinch Events only. " +
+                $"Configure either {nameof(SinchSmsConfiguration)}.{nameof(SinchSmsConfiguration.Region)} or " +
+                $"{nameof(SinchSmsConfiguration)}.{nameof(SinchSmsConfiguration.ServicePlanIdConfiguration)} to use SMS API operations.");
```
JPPortier: We should point to the exact root cause of what is missing to help end user to configure properly configuration to be used when using SMS.
- not have added value to talk about Sinch Events only. If there is an issue about config it could (should) not being related to Sinch Events usage at all
- Instead of having end user to detect between "or" possible missing values: highlight which one(s) are missing

tests/Sinch.Tests/Sms/SmsClientTests.cs
```diff
@@ -64,9 +65,14 @@ public void ResolveSmsUrl_ThrowsWhenRegionNotSet()
                     ProjectId = "project-id"
                 }
             });
-            var act = () => client.Sms;
+
+            client.Sms.SinchEvents.Should().NotBeNull();
```
JPPortier: Let say this is not related to the current test tile and is already covered here: https://github.com/sinch/sinch-sdk-dotnet/blob/d306e325794d0154fc5c26d51cc070500cc5773f/tests/Sinch.Tests/Sms/SmsClientTests.cs#L107-L125


```suggestion
            
```

MIGRATION_GUIDE.md
```diff
@@ -522,32 +521,124 @@ var bin = JsonSerializer.Deserialize<IncomingBinarySms>(json);
...
+{
+    CallbackUrl = new Uri("https://my-server.com/sms-events")
+};
+```
+
+**After:**
+```csharp
+var request = new SendTextBatchRequest
+{
+    EventDestinationTarget = new Uri("https://my-server.com/sms-events")
```
JPPortier: This is no longer an `Uri` right ?

JPPortier (approved):
🚀

PR #203, opened 2026-05-12
Voice - Redesign of webhooks and callback configuration

rpredescu-sinch:
## Summary

Voice to "Event Destinations'" and "Sinch Events" terminology redesign.

## What it is in this PR

- `Sinch.Voice.Hooks`  namespace renamed to `Sinch.Voice.SinchEvents`
- SinchVoiceClient lazily check the credentials when executing a Voice action


src/Sinch/Voice/SinchEvents/VoiceSinchEvent.cs
```diff
@@ -0,0 +1,15 @@
+using System.Text.Json.Serialization;
+
+namespace Sinch.Voice.SinchEvents
+{
+    /// <summary>
+    ///     Base type for all Voice Sinch Events.
+    /// </summary>
+    [JsonConverter(typeof(VoiceSinchEventConverter))]
+    public abstract class VoiceSinchEvent
```
JPPortier: It should, like for other domains, implement a `IVoiceSinchEvent` right ?

src/Sinch/Voice/SinchEvents/IVoiceSinchEvents.cs
```diff
@@ -0,0 +1,57 @@
...
+    ///     and serialize event responses.
+    /// </summary>
+    public interface IVoiceSinchEvents
+    {
+        /// <summary>
+        ///     Parse a Voice Sinch Event from a raw JSON string.
+        /// </summary>
+        /// <param name="json">Raw JSON string of the incoming event payload.</param>
+        /// <returns>The parsed <see cref="VoiceSinchEvent" />.</returns>
+        VoiceSinchEvent ParseEvent(string json);
```
JPPortier:
```suggestion
        IVoiceSinchEvent ParseEvent(string json);
```

src/Sinch/Voice/SinchEvents/VoiceSinchEvents.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        /// <inheritdoc />
+        public bool ValidateAuthenticationHeader(
+            HttpMethod method,
+            string path,
+            IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers,
+            string body)
+        {
+            EnsureCredentials();
```
JPPortier: Is there a reason to not having follow same pattern as in previous PR for Verification ? (https://github.com/sinch/sinch-sdk-dotnet/blob/feature/DEVEXP-1311-voice-event-destinations-and-sinch-events-v2/src/Sinch/Verification/SinchEvents/VerificationSinchEvents.cs#L54)

JPPortier (dismissed):
This PR should also cover "callbacks" as "Event Destinations": https://github.com/sinch/sinch-sdk-dotnet/blob/feature/DEVEXP-1311-voice-event-destinations-and-sinch-events-v2/src/Sinch/Voice/Applications/Applications.cs#L44-L60

src/Sinch/Voice/SinchEvents/VoiceSinchEvent.cs
rpredescu-sinch: Good point, but there is a trade off which needs to be acknowledged.

`EventType Event` is a natural contract: it is shared by every concrete Voice event type and belongs on the base type regardless of whether that is an **interface** or an **abstract** class.

- An `abstract class` can declare `internal abstract EventType Event`, the discriminator is used internally for **deserialization** and `ToString()` but never becomes part of the public API surface.
- An `interface` forces every member to be public and once `EventType Event` is on a public interface, it is locked until the next major version.

 If you prefer the interface for consistency with `INumbersSinchEvent` and `ISmsSinchEvent`, we can make that change, but just wanted to make sure the trade-off is a conscious decision.

### Note

For **Verification** we have the abstract class, and for a different reason than **Voice**.

VerificationSinchEvent holds actual shared data properties (`Id, Event, Method, Identity, Reference, Custom`) all with `[JsonPropertyName]` mappings, and all present on every concrete event type (`VerificationStartEvent, VerificationResultEvent, VerificationSmsDeliveredEvent`). An abstract class lets us define them once with their JSON attributes in one place.

If we used an interface instead, those properties would have to be declared twice: once on the `interface` (no [JsonPropertyName]) and once on every concrete class (with [JsonPropertyName]). That duplication is exactly what the abstract class avoids.

This is different from the Voice case, where the base type only held a single discriminator field (`EventType Event`) that varies per concrete type anyway.

src/Sinch/Voice/SinchEvents/IVoiceSinchEvents.cs
rpredescu-sinch: See my comment here: https://github.com/sinch/sinch-sdk-dotnet/pull/203/changes#r3241905024

src/Sinch/Voice/SinchEvents/VoiceSinchEvents.cs
rpredescu-sinch: No special reason, I have aligned credential guard with Verification pattern.

src/Sinch/Voice/SinchEvents/VoiceSinchEvent.cs
JPPortier: According to discussion we will go to define
-  an empty `IVoiceSinchEvent`
- define `VoiceSinchEvent : IVoiceSinchEvent`
- programmatic APIs will use the `IVoiceSinchEvent` (e.g. ParseEvent)
- then we will have a common pattern across all domains

examples/templates/sinch-events/Voice/VoiceSinchEventsController.cs
```diff
@@ -0,0 +1,65 @@
+using Microsoft.AspNetCore.Mvc;
+using Sinch.Voice.SinchEvents;
+
+namespace SinchEvents.Template.Voice;
+
+/// <summary>
+/// Example controller for receiving Voice Sinch Events.
+/// </summary>
+[ApiController]
+[VoiceSinchEvent(requireAuthentication: false)]
```
asein-sinch, line 10: Rather than `requireAuthentication`, it would be clearer to name this `requireSignatureValidation`

examples/templates/sinch-events/Voice/VoiceSinchEventsController.cs
```diff
@@ -0,0 +1,65 @@
...
+        var body = HttpContext.Items[SinchEventsConstants.BodyItemKey] as string;
+        var voiceEvent = _voiceSinchEvents.ParseEvent(body!);
+
+        SinchEventResponse? response = voiceEvent switch
+        {
+            IncomingCallEvent ice => _businessLogic.IncomingCallEvent(ice),
+            AnsweredCallEvent ace => _businessLogic.AnsweredCallEvent(ace),
+            DisconnectedCallEvent dice => HandleDisconnected(dice),
+            PromptInputEvent pie => _businessLogic.PromptInputEvent(pie),
+            NotificationEvent notify => HandleNotification(notify),
```
asein-sinch, line 40: I'd like to see the same pattern everywhere. It's true that some events expect a response and some others not.
In this code, it's confusing just because the some functions in the business logic layer are returning `void`. Instead, it would be better to call the business logic in ALL the `switch` and to return something in ALL methods. When the event type doesn't require a SVAML response (DICE, Notification), return `null`, but for the others (ACE, ICE, PIE), let's go the extra mile and return a real SVAML response instead of `null` in the current code.

src/Sinch/Verification/SinchEvents/VerificationEvent.cs
asein-sinch, line 1: Is it a result of a bad merge? This class is abstract, doesn't belong to the Voice domain and is used nowhere in the code except in the README which looks fishy.

src/Sinch/Voice/Applications/UpdateEventDestinations/UpdateEventDestinationsRequest.cs
```diff
@@ -1,29 +1,24 @@
...
-
         public required string ApplicationKey { get; set; }
 
         /// <summary>
-        ///     Gets or Sets Url
+        ///     The event destination URLs to configure.
         /// </summary>
-        public CallbackUrls? Url { get; set; }
-
+        public EventDestinationTarget? Url { get; set; }
```
asein-sinch, line 15: Should stay plural as the underlying schemas contains 2 URLs

src/Sinch/Voice/Destinations/IStreamDestination.cs
```diff
@@ -0,0 +1,8 @@
+namespace Sinch.Voice.Destinations
+{
+    /// <summary>
+    ///     Marker interface for destinations valid in streaming SVAML actions (e.g. <c>connectStream</c>):
+    ///     <see cref="DestinationWebSocket"/>.
+    /// </summary>
+    public interface IStreamDestination : IDestination { }
```
asein-sinch, line 7: Why do you need an extra layer of interface for this destination?

src/Sinch/Voice/Destinations/DestinationWebSocket.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace Sinch.Voice.Destinations
+{
+    /// <summary>A WebSocket stream endpoint used with the <c>connectStream</c> SVAML action.</summary>
+    public sealed class DestinationWebSocket : IStreamDestination
+    {
+        /// <inheritdoc />
+        [JsonPropertyName("type")]
+        public DestinationType Type => DestinationType.Websocket;
+
+        /// <inheritdoc />
```
asein-sinch, line 12: Shouldn't you override the description in this implementation class?

src/Sinch/Voice/Destinations/SinchEventDestinationConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+        }
+
+        public override void Write(Utf8JsonWriter writer, ISinchEventDestination value, JsonSerializerOptions options)
+        {
+            switch (value)
+            {
+                case DestinationPstn pstn: JsonSerializer.Serialize(writer, pstn, options); break;
+                case DestinationMxp mxp: JsonSerializer.Serialize(writer, mxp, options); break;
+                case DestinationSip sip: JsonSerializer.Serialize(writer, sip, options); break;
+                case DestinationDid did: JsonSerializer.Serialize(writer, did, options); break;
```
asein-sinch, line 34: Wouldn't it be more readable to combine all 4 cases instead of repeating the same code?

src/Sinch/Voice/Calls/Manage/ManageWithCallLegRequest.cs
```diff
@@ -1,7 +1,7 @@
 using System.Collections.Generic;
 using System.Text;
-using Sinch.Voice.Calls.Actions;
-using Sinch.Voice.Calls.Instructions;
+using Sinch.Voice.Svaml.Actions;
+using Sinch.Voice.Svaml.Instructions;
```
asein-sinch, line 4: 👍

src/Sinch/Voice/Calls/Call.cs
```diff
@@ -11,12 +12,12 @@ public sealed class Call
         /// <summary>
         ///     Contains the caller information.
         /// </summary>
-        public Destination? From { get; set; }
+        public ICalloutDestination? From { get; set; }
```
asein-sinch, line 15: Even if it's the same class, it feels odd to read "Destination" for an origin

src/Sinch/Voice/Svaml/Actions/ConnectPstn.cs
```diff
@@ -102,7 +102,7 @@ public override string ToString()
             sb.Append("  MaxDuration: ").Append(MaxDuration).Append("\n");
             sb.Append("  DialTimeout: ").Append(DialTimeout).Append("\n");
             sb.Append("  Cli: ").Append(Cli).Append("\n");
-            sb.Append("  SuppressCallbacks: ").Append(SuppressCallbacks).Append("\n");
+            sb.Append("  SuppressCallbacks: ").Append(SuppressSinchEvents).Append("\n");
```
matsk-sinch, line 105: Should the label be `SuppressSinchEvents` too?

src/Sinch/Voice/Svaml/Actions/ConnectSip.cs
```diff
@@ -95,7 +95,7 @@ public override string ToString()
             sb.Append("  MaxDuration: ").Append(MaxDuration).Append("\n");
             sb.Append("  Cli: ").Append(Cli).Append("\n");
             sb.Append("  Transport: ").Append(Transport).Append("\n");
-            sb.Append("  SuppressCallbacks: ").Append(SuppressCallbacks).Append("\n");
+            sb.Append("  SuppressCallbacks: ").Append(SuppressSinchEvents).Append("\n");
```
matsk-sinch, line 98: Should this label be updated?

src/Sinch/Voice/Destinations/ISinchEventDestination.cs
```diff
@@ -0,0 +1,11 @@
+using System.Text.Json.Serialization;
+
+namespace Sinch.Voice.Destinations
+{
+    /// <summary>
+    ///     Marker interface for destinations valid in Voice Sinch Event payloads (ICE / DICE <c>to</c> field):
+    ///     <see cref="DestinationPstn"/>, <see cref="DestinationMxp"/>, <see cref="DestinationSip"/>, <see cref="DestinationDid"/>.
+    /// </summary>
+    [JsonConverter(typeof(SinchEventDestinationConverter))]
+    public interface ISinchEventDestination : IDestination { }
```
asein-sinch, line 10: The name of this interface is confusing because we are not talking about the "SinchEvent destination", the new name for this redesign for the callback URLs, but really about the call destination

src/Sinch/Voice/SinchEvents/DisconnectedCallEvent.cs
```diff
@@ -81,7 +52,7 @@ public sealed class DisconnectedCallEvent : IVoiceEvent
         ///     Gets or Sets To
         /// </summary>
         [JsonPropertyName("to")]
-        public To? To { get; set; }
+        public ISinchEventDestination? To { get; set; }
```
asein-sinch, line 55: This is where I started to be confused: "To" was more meaningful as a classname to represent who/what is being called

src/Sinch/Voice/SinchEvents/DisconnectedCallEvent.cs
```diff
@@ -127,19 +85,14 @@ public override string ToString()
             var sb = new StringBuilder();
             sb.Append("class DiceRequest {\n");
             sb.Append("  Event: ").Append(Event).Append("\n");
```
asein-sinch, line 87: What is the difference with 
```
sb.Append($"  {nameof(Event)}: ").Append(Event).Append('\n');
```

src/Sinch/Voice/Hooks/DisconnectedCallEvent.cs
```diff
@@ -127,19 +85,14 @@ public override string ToString()
             var sb = new StringBuilder();
             sb.Append("class DiceRequest {\n");
             sb.Append("  Event: ").Append(Event).Append("\n");
-            sb.Append("  CallId: ").Append(CallId).Append("\n");
-            sb.Append("  Timestamp: ").Append(Timestamp).Append("\n");
             sb.Append("  Reason: ").Append(Reason).Append("\n");
             sb.Append("  Result: ").Append(Result).Append("\n");
-            sb.Append("  Version: ").Append(Version).Append("\n");
-            sb.Append("  Custom: ").Append(Custom).Append("\n");
```
asein-sinch, line 135: These properties are being inherited. Why not printing them in the actual class?

src/Sinch/Voice/SinchEvents/SinchEventResponse.cs
```diff
@@ -1,14 +1,14 @@
...
-using Sinch.Voice.Calls.Actions;
-using Sinch.Voice.Calls.Instructions;
+using Sinch.Voice.Svaml.Actions;
+using Sinch.Voice.Svaml.Instructions;
 
-namespace Sinch.Voice.Hooks
+namespace Sinch.Voice.SinchEvents
 {
     /// <summary>
     ///     The Incoming Call Event (ICE) or The Answered Call Event (ACE) requires a valid SVAML object in response.
```
asein-sinch, line 9: PIE too if I'm not mistaken

src/Sinch/Voice/Svaml/Actions/ConnectPstn.cs
```diff
@@ -59,7 +59,7 @@ public sealed class ConnectPstn : IAction
         ///     <see href="https://developers.sinch.com/docs/voice/api-reference/voice/voice/tag/Callbacks/#tag/Callbacks/operation/dice">DICE</see> callbacks for the call.
```
asein-sinch, line 59: Note, the term "callback" is still mentioned in the description

src/Sinch/Voice/SinchEvents/VoiceSinchEvent.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    ///     Base type for all Voice Sinch Events.
+    /// </summary>
+    public abstract class VoiceSinchEvent : IVoiceSinchEvent
+    {
+        [JsonPropertyName("event")]
+        [JsonInclude]
+        internal abstract EventType Event { get; set; }
+
+        /// <inheritdoc />
```
asein-sinch, line 14: Shouldn't it be a real description?

src/Sinch/Voice/SinchVoiceClient.cs
```diff
@@ -41,114 +36,88 @@ public interface ISinchVoiceClient
...
+            ApplicationSignedAuth? auth = null;
+            Http? http = null;
+
+            if (config != null)
+            {
+                if (string.IsNullOrEmpty(config.AppKey))
+                    throw new ArgumentNullException(nameof(config.AppKey), "The value should be present");
+
+                if (string.IsNullOrEmpty(config.AppSecret))
+                    throw new ArgumentNullException(nameof(config.AppSecret), "The value should be present");
```
asein-sinch, line 75: I believe there is a ticket to address this, right?

tests/Sinch.Tests.Features/Voice/Utils.cs
```diff
@@ -19,7 +19,7 @@ public class Utils
                     VoiceConfiguration = new SinchVoiceConfiguration()
                     {
                         AppKey = "appKey",
-                        AppSecret = "BeIukql3pTKJ8RGL5zo0DA=="
+                        AppSecret = "YXBwU2VjcmV0"
```
asein-sinch, line 22: How is it possible that the tests are still working / were working with a different value of the secret?

tests/Sinch.Tests/Voice/VoiceClientTests.cs
```diff
@@ -115,15 +115,28 @@ public void ThrowIfVoiceCredentialsAreMissing(VoiceCredentialsMissingTestCaseDat
         }
 
         [Fact]
-        public void ThrowIfVoiceConfigIsNull()
+        public void SinchEvents_ParseEvent_DoesNotRequireConfiguration()
```
asein-sinch, line 118: Thanks for fixing this 👍

MIGRATION_GUIDE.md
```diff
@@ -1358,11 +1368,33 @@ Version 2.*:
 EventType.DeprovisioningFromCampaign
 ```
 
+<<<<<<< HEAD
```
asein-sinch, line 1371: Remove the conflict markers

src/Sinch/Voice/SinchEvents/VoiceSinchEvents.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                throw new InvalidOperationException("Deserialization of Voice Sinch Event failed");
+            }
+
+            return result;
+        }
+
+        /// <inheritdoc />
+        public bool ValidateAuthenticationHeader(
+            HttpMethod method,
```
asein-sinch, line 43: Design question: as the event will always be sent as a POST, shouldn't we hardcode this value in the algorithm rather than asking the user to add it each time?

